PR #1, opened 2015-03-18
Error check to pdsh function and other works to verify the deploying process

xuechendi:
Four commit here:

1. add error check in pdsh function, so pdsh will exit when stderr is not none.
2. add osd device field in yaml, user can specify osd devices there, if none, will follow the original way using /dev/disk/by-partlabel/osd-device-%s-data
3. skip error check when calling ceph-osd, ceph-mon command
4. change pdcp ceph.conf to all nodes to scp

### DIFF
--- a/common.py
+++ b/common.py
@@ -3,40 +3,46 @@ import subprocess
 import time
 import os
 import errno
+import sys
 
-def pdsh(nodes, command):
+def pdsh(nodes, command, force=False):
     args = ['pdsh', '-R', 'ssh', '-w', nodes, command]
     print('pdsh: %s' % args)
-    return subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+    stdout, stderr = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True).communicate()
+    if force:
+        return [stdout, stderr]
+    if stderr:
+        print "[ERROR]:"+stderr+"\n"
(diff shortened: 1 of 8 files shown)


common.py
```diff
@@ -3,40 +3,46 @@
...
 
-def pdsh(nodes, command):
+def pdsh(nodes, command, force=False):
     args = ['pdsh', '-R', 'ssh', '-w', nodes, command]
     print('pdsh: %s' % args)
-    return subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+    stdout, stderr = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True).communicate()
+    if force:
+        return [stdout, stderr]
+    if stderr:
```
koder-ua: Having something in stderr doesn't mean, that command fail

bengland2:
that's why I checked Popen.returncode in my pull request.

bengland2:
I do like idea of specifying block devices in yaml instead of with partition name convention though!  Can we separate out that part and get that merged?

bengland2:
you have a point about using scp in at least one case,  rpdcp  did not pull results back to test driver for me, but I think it can push files out from test driver to other systems just fine.  We could do scps in parallel to copy per-host results to different subdirectories on test driver host.

bengland2:
I think the error checking part of this pull request is now fixed, or at least mostly fixed, more error checking wlil be added over time, see merged PRs #110 and #107 .

PR #20, opened 2015-08-03
cbt.py pythonization, add logging module, replace print in cbt.py with logging

koder-ua:


markhpc:
Hi,

Would you please separate out the logging changes and the pythonization changes into the seperate pull requests?

Thanks!
Mark

koder-ua:
Will try

PR #22, opened 2015-08-05
check status returned by pdsh where appropriate

bengland2:
I have tested CBT with use_existing=True, haven't yet tested the other way.  This was the least disruptive patch I could come up with that does check pdsh status.

### DIFF
--- a/common.py
+++ b/common.py
@@ -4,37 +4,75 @@ import time
 import os
 import errno
 
-def pdsh(nodes, command):
-    args = ['pdsh', '-R', 'ssh', '-w', nodes, command]
-    print('pdsh: %s' % args)
-    return subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+# this class overrides the communicate() method to check the return code and
+# throw an exception if return code is not OK
+
+class checked_Popen:
+    UNINIT=-720
+    OK=0
+    # if you really don't care and want old unchecked behavior, set this to True, 
+    # but consider setting "continue_if_error=True" in individual common.pdsh call instead
+    turn_off_checking=False  
+    loglevel = True
+    def __init__(self, args, continue_if_error=False):
+        self.args = args[:]
+        self.myrtncode = self.UNINIT
+	self.continue_if_error = continue_if_error
+        self.popen_obj = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+        if self.loglevel: print('%s: %s' % (time.ctime(), ' '.join(args)))
+
+    def __str__(self):
+       return 'checked_Popen args=%s continue_if_error=%s rtncode=%d'%(str(self.args), str(self.continue_if_error), self.myrtncode)
+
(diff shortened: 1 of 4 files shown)


markhpc:
Hi Ben, please rebase on master

common.py
```diff
@@ -4,37 +4,75 @@
 import os
 import errno
 
-def pdsh(nodes, command):
-    args = ['pdsh', '-R', 'ssh', '-w', nodes, command]
-    print('pdsh: %s' % args)
-    return subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+# this class overrides the communicate() method to check the return code and
+# throw an exception if return code is not OK
```
alfredodeza: that is what `Popen.check_call` is for: https://docs.python.org/2/library/subprocess.html#subprocess.check_call

"Run command with arguments. Wait for command to complete. If the return code was zero then return, otherwise raise CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute."

bengland2:
I took a PTO day today but will do so by end of Monday.

----- Original Message -----

common.py
bengland2: Alfredo (Deza), 

as you say, Popen.check_call() method is intended to both launch the command and check the exit status.  That was the first change I considered.  

However, one of the hidden attributes of the existing pdsh() call is that it starts a command running (with Popen() constructor) but doesn't block on its completion, so that you can start multiple pdsh sessions going in parallel.  This is a "poor man's multi-threading", but this capability would be lost if we started using check_call().  

So why do we need parallel pdsh sessions?  One reason is that radosbench can run more than one process per client host (used in benchmarks/radosbench.py) .  There may be other instances where parallel pdsh sessions are used.  I guess that Mark achieves full parallelism in rados bench by having multiple rados bench processes per host on multiple client hosts, with each rados bench process targeted at a different Ceph storage pool.  See where it launches rados bench pdsh commands, and then calls p.wait() on each.  

Another hidden attribute of existing pdsh() call is the ability to ignore errors and not throw an exception if issuing a command whose execution is optional.   The hippocratic oath is "first do no harm"!  So in making this change to strengthen checking, I don't want to have the side effect that tests which would otherwise complete no longer finish.   That's why in my pull request I define a new pdsh argument continue_if_error that defaults to False, but can be set to True by the caller if the intent is that the benchmark should not terminate if the command fails (but a warning will appear).   For example, if collectl or blktrace/seekwatcher command fails, do we abort the entire test?  When running a long series of tests, it's often very painful to have something terminate completely in the middle of the sequence, when it would be possible to get at least partial results and learn something from a continuation.

So in my pull request (soon to be rebased) I preserve the functioning of the existing code by doing the status check (if wanted) after Popen.communicate() method had completed, an alternative that is specified by Python documentation for Popen.returncode.  Generally where possible I prefer small, incremental changes that perform one enhancement at a time, and can be easily verified independently of one another.  Make sense?

-ben



----- Original Message -----

bengland2:
I'm closing this pull request, it's superceded by https://github.com/ceph/cbt/pull/27 which incorporates Konstantin's print->logger changes

PR #26, opened 2015-08-10
Pythonize benchmarkfactory.py cbt.py monitoring.py settings.py; Add UT for benchmarkfactory.py

koder-ua:


markhpc:
LGTM, in testing

PR #27, opened 2015-08-10
use common.CheckedPopen class, additional changes for use_existing: F…

bengland2:
This has been rebased against Konstantin's print->logging commit.  This does not yet completely run in my environment, it currently works with use_existing: True and it works with use_existing: False up to the point where it tries to start ceph-mon.  I'll continue to work on resolving problems until it runs clean but thought others might want to see what's done, in light of Konstantin's latest pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,50 @@
 
 ## INTRODUCTION
 
-CBT is a python tool for building a ceph cluster and running benchmarks against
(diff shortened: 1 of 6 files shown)


bengland2:
I'm withdrawing this pull request in favor of another one with only 1 commit including rebase

PR #29, opened 2015-08-10
set_osd_param: add support for use_existing use case

### LINKED ISSUE #25: Readahead for use_existing
When using use_existing, readahead is not set on the OSDs because they do not have the labels CBT expects. We should probably determine which devices are mounted in /var/lib/osd/* and then set their readahead values.

mmgaggle:
set_osd_param used a command that required that the OSDs use the CBT
style labels, which is unlikely during a use_existing use case. This
changes the command to modify the device settings based on devices
that have a filesystem mounted in /var/lib/ceph/osd instead of by
label. Closes #25

### DIFF
--- a/cluster/ceph.py
+++ b/cluster/ceph.py
@@ -287,7 +287,7 @@ class Ceph(Cluster):
         common.pdsh(settings.getnodes('osds'), 'find /var/run/ceph/*.asok -maxdepth 1 -exec sudo ceph --admin-daemon {} dump_historic_ops \; > %s/historic_ops.out' % run_dir).communicate()
 
     def set_osd_param(self, param, value):
-        common.pdsh(settings.getnodes('osds'), 'find /dev/disk/by-partlabel/osd-device-*data -exec readlink {} \; | cut -d"/" -f 3 | sed "s/[0-9]$//" | xargs -I{} sudo sh -c "echo %s > /sys/block/\'{}\'/queue/%s"' % (value, param))
+        common.pdsh(settings.getnodes('osds'), 'mount|grep "/var/lib/ceph/osd"|sed -e "s/\s.*$//" -e s"/.$//g" -e s"/\/dev\///g" | xargs -I{} sudo sh -c "echo %s > /sys/block/\'{}\'/queue/%s"' % (value, param))
 
 
     def __str__(self):


markhpc:
So I guess one question I have here is do we want CBT to modify block device settings on existing clusters?  Kind of a goal I've tried to have with use_existing is to not modify an existing cluster.  Changing the block device settings theoretically should be fine I suppose as a reboot should restore the default behavior, though it makes me a little nervous.  If we do adopt this, I think we need to change the way device detection in CBT works so that we aren't limited to either the /var or the <cbt_tmp>/mnt/* symlinks.

mmgaggle:
I'm ok with not having use_existing modify device settings. Perhaps instead we should skip past osd_ra code paths if use_existing is set, and also not have osd_ra in the path of the results directory. My main objection to the current behavior is having to explain to people that use use_existing, that osd_ra is required in the yaml, but not actually used, and that the osd_ra in the results path is meaningless.

markhpc:
Fair enough!  That's pretty asinine behavior. :)

Mark

mmgaggle:
I think something as easy as this could do it without breaking runs where you have an array of osd_ra values. Thoughts?

```
diff --git a/benchmark/benchmark.py b/benchmark/benchmark.py
index 2d88f4c..b73dbb9 100644
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -36,9 +36,12 @@ class Benchmark(object):
         common.make_remote_dir(self.run_dir)

     def run(self):
-        logger.info('Setting OSD Read Ahead to: %s', self.osd_ra)
         if self.osd_ra:
+            logger.info('Setting OSD Read Ahead to: %s', self.osd_ra)
             self.cluster.set_osd_param('read_ahead_kb', self.osd_ra)
+        else
+            self.osd_ra = 'na'
+
         logger.debug('Cleaning existing temporary run directory: %s', self.run_dir)
         common.pdsh(settings.getnodes('clients', 'osds', 'mons', 'rgws'), 'sudo rm -rf %s' % self.run_dir).communicate()
         if self.valgrind is not None:
```

markhpc:
lgtm!  Want to update the PR or should I change it?

mmgaggle:
I'll send a new PR, thanks!

PR #49, opened 2015-08-26
add tox files

koder-ua:


markhpc:
I haven't used tox, so I'll trust these are reasonable and we'll fix later if they aren't. :)  Added to my test branch.

PR #51, opened 2015-08-27
do mkfs in parallel (except for tmpfs, zfs)

bengland2:
This particular pull request was just tested against examples/wip-mark-testing using a VM with 1 mon and 1 OSD, but has also been used on a 232-OSD Ceph cluster with dramatic speedup, 80 seconds to initialize all 232 2.7-TB disk drives, 1.4 GB/sec/server peak throughput.


markhpc:
Awesome, if it's passing the tests (as anemic as they are right now) let's merge.

PR #52, opened 2015-09-01
radosbench benchmark uses 1 storage pool, can revert to 1 pool per rados proc

bengland2:
This is more real-world use case since pools are typically shared by many rados clients, and it enables use of larger PG counts for this one pool for scalability testing.  Tested with example/wip-mark-testing, also with subset setting radosbench param "pool_per_proc" to True, which reproduces prior behavior of a separate storage pool per rados bench process.


markhpc:
This is a good change, but does mean that there will be fewer pools used and thus fewer total PGs unless the yaml configuration file is changed.

PR #55, opened 2015-09-09
error-check CBT subprocess return status

bengland2:
this replaces pull request 27.  It runs with example/wip-mark-testing, with one exception: I can't get all the tests to pass in a single CBT run.  I am working on that separately.

The CheckedPopen class is what the common pdsh,pdcp, and other functions return now, instead of a subprocess.Popen object.  This allows us to override the subprocess.Popen.communicate() method to do error checking, minimizing changes to CBT code.  It throws an exception if the subprocess exit status is not 0.

in a couple of places we change p.wait() to p.communicate() and just ignore the outputs.

In some places we use the optional continue_if_error parameter to the CheckedPopen constructor so that it will avoid stopping CBT if a failure occurs, but it will log a warning message instead.

### DIFF
--- a/benchmark/cephtestrados.py
+++ b/benchmark/cephtestrados.py
@@ -124,7 +124,7 @@ class CephTestRados(Benchmark):
         monitoring.stop()
 
     def recovery_callback(self): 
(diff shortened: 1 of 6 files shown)


bengland2:
retiring this one, it is replaced by PR 107, which is similar but less radical in scope.

PR #56, opened 2015-09-10
Log color suppression

bengland2:
Suppress color output in log messages unless stdout is a tty.  Also change CRITICAL to RED and INFO to GREEN.

Please ignore the first two commits, the 2nd undoes the first on the master branch, ouch.  The 3rd one is what I intended.

### DIFF
--- a/log_support.py
+++ b/log_support.py
@@ -1,7 +1,8 @@
 import logging
-
+import os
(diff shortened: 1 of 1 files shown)


bengland2:
will clean up and re-submit

PR #57, opened 2015-09-10
Control loglevel

bengland2:
This commit lets the user specify logging level using LOGLEVEL environment variable.  If LOGLEVEL is not defined, the default log level of logging.INFO is used instead of logging.DEBUG.

please ignore the first 2 commits, the 2nd one cancels the first commit to the master branch, sorry again.

### DIFF
--- a/log_support.py
+++ b/log_support.py
@@ -1,7 +1,8 @@
 import logging
-
+import os
 import yaml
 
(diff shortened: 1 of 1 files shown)


bengland2:
will clean up and resubmit

PR #58, opened 2015-09-14
Add run/output directory for varying read percentages

ommoreno:
Will add "readmix-n" to the directory structure, where n is the read percentage. This allows for multiple rwmixread values in the cbt yaml configuration file.

### DIFF
--- a/benchmark/librbdfio.py
+++ b/benchmark/librbdfio.py
@@ -32,12 +32,14 @@ class LibrbdFio(Benchmark):
         self.log_avg_msec = config.get('log_avg_msec', None)
 #        self.ioengine = config.get('ioengine', 'libaio')
         self.op_size = config.get('op_size', 4194304)
+        self.bssplit = config.get('bssplit', None)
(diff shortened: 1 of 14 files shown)


markhpc:
Hi, I like the idea, but it will likely break folks existing parsing scripts. I think this showcases how badly the existing way we try to store test run results scales.  I think instead we should change CBT to store results in a UUID named directory (distinct for each test run) and then put the parameters associated with that test in a flat file in the directory.  We could then write tools (and possibly caches results in a database) to search query for the results.  I think this is the only way we'll be able to handle all of the different permutations of tests that are now possible.

ommoreno:
I agree. That is a good idea and I could definitely see that being implemented in the future. This was just a fix I made for myself and thought it might be useful to others who want to automate running different tests.

markhpc:
Closing this for now, hopefully we'll have an intern this summer that can help out with the directory structure work that will make features like this more easy to integrate!

Issue #43, opened 2015-08-11
To create a cosbench yaml example


prallabh:
Can be on the lines of from a cosbench profile:

<?xml version="1.0" encoding="UTF-8" ?> 
- <workload name="64K-Write-Workload">
  <auth type="swauth" config="username=johndoe:swift;password=asdf;auth_url=http://192.168.1.1:8686/auth/v1.0" /> 
  <storage type="swift" config="" /> 
- <workflow>
- <workstage name="init">
  <work type="init" workers="1024" config="containers=r(1,100)" /> 
  </workstage>
- <workstage name="prepare-write">
  <work type="prepare" workers="1024" config="containers=r(1,100);objects=r(1,1000);sizes=c(1)MB" /> 
  </workstage>
  </workflow>
  </workload>

PR #63, opened 2015-11-12
Update cosbench.py to add cosbench user/subuser if use_existing is False

wadeholler:
Update cosbench.py to add cosbench user/subuser if use_existing is False

--There may be more elegant ways to update radosgw with the cosbench:operator sub user but this follows the cosbench documentation.
--Only update is use_existing is False

Issue #64 

While we are at it.  Add a simple cosbench example per Issue #43

### DIFF
--- a/benchmark/cosbench.py
+++ b/benchmark/cosbench.py
@@ -28,6 +28,8 @@ class Cosbench(Benchmark):
         self.objects = config["objects_max"]
         self.mode = config["mode"]
         self.user = settings.cluster.get('user')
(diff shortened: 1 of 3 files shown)


markhpc:
I'm going to merge this for now, but I think we should try to keep rados stuff out of the benchmarks, so we may want to move this into the ceph cluster class and have the benchmark use some kind of generic interface for creating users and whatnot.  The eventual goal would be that we could create a swift backend in addition to ceph and use the same benchmark.

PR #68, opened 2015-11-19
Update rbdfio cleanup

wadeholler:
Devices could be left in a stale state on the clients because there was no cleanup in benchmark rbdfio. This could result in later, other benchmark runs failing during their sync step on the clients. 

Updated with steps:
umount
rbd unmap
rbd rm

### DIFF
--- a/benchmark/cephfsfio.py
+++ b/benchmark/cephfsfio.py
@@ -1,0 +1,169 @@
+import subprocess
+import common
+import settings
+import monitoring
+import os
+import time
+import logging
+
+from benchmark import Benchmark
+
+logger = logging.getLogger("cbt")
+
+class CephFsFio(Benchmark):
+
+    def __init__(self, cluster, config):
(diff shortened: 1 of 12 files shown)


markhpc:
Hrm, I think perhaps we should have a flag here.  One of the things that we've used CBT for in the past is to let a set of tests run, look at the results, and then run new tests by hand after the tests complete to track down performance anomalies.  Perhaps it makes sense though to have the default behavior be to cleanup the rbd mount points to avoid this kind of issue and only leave the mount points enabled if some kind of "no_cleanup" flag is passed or something.

mmgaggle:
In the librbdfio benchmark we have use_existing_volumes. The problem with use_existing_volumes is it doesn't create the volumes if they already exist. Ideally there would be an option where it creates and initializes the volumes if they don't exist, but skips those steps if they do, while also not cleaning up.

PR #85, opened 2016-02-15
Remove potentially dangerous mkfs option from samples

jrosenboom:
As using a large directory block size on XFS can turn out to have pretty
bad effects on systems under load (see [1]), do not suggest using these
as a default in the examples.

[1] http://oss.sgi.com/pipermail/xfs/2016-February/047020.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ cluster:
       a: "127.0.0.1:6789"
   osds_per_node: 1
   fs: 'xfs'
-  mkfs_opts: '-f -i size=2048 -n size=64k'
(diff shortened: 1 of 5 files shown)


markhpc:
FWIW, this option originally was included based on a recommendation to test by Christoph Hellwig back in 2012:

http://permalink.gmane.org/gmane.comp.file-systems.ceph.devel/11568

We had a number of folks in the community and Inktank trying out different XFS options to see if they improved performance, and the ones included in the examples were more or less the result of what people were generally having the most success with at the time.  Based on David's comments it sounds like removing these from the examples is a good idea though.

PR #90, opened 2016-02-24
Added hotfix for benchmarks breaking if osd_ra is not set in the yaml config file

roclark:
If the osd_ra parameter is not set in the yaml config file, the default value given to osd_ra is None. When not specified, the benchmark classes will break when trying to parse an int from None. For example, the following line is in the KvmRbdFio class (as well as in most other benchmark classes):
`self.run_dir = '%s/osd_ra-%08d/client_ra-%08d/op_size-%08d/concurrent_procs-%03d/iodepth-%03d/%s' % (self.run_dir, int(self.osd_ra), int(self.client_ra), int(self.op_size), int(self.total_procs), int(self.iodepth), self.mode)`

Once the above line is executed, it will attempt to parse an int from `None` and CBT will break with a `TypeError`. This hotfix checks for a `read_ahead_kb` value in the cluster and uses that value for osd_ra if not specified.

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -20,6 +20,9 @@ class Benchmark(object):
         self.cmd_path_full = '' 
         if self.valgrind is not None:
             self.cmd_path_full = common.setup_valgrind(self.valgrind, self.getclass(), self.run_dir)
+        if self.osd_ra is None:
+            self.osd_ra = common.get_osd_ra()
+            self.osd_ra_changed = False
 
 
     def getclass(self):
@@ -37,7 +40,7 @@ class Benchmark(object):
 
     def run(self):
         logger.info('Setting OSD Read Ahead to: %s', self.osd_ra)
-        if self.osd_ra:
+        if self.osd_ra and self.osd_ra_changed is not False:
             self.cluster.set_osd_param('read_ahead_kb', self.osd_ra)
         logger.debug('Cleaning existing temporary run directory: %s', self.run_dir)
         common.pdsh(settings.getnodes('clients', 'osds', 'mons', 'rgws'), 'sudo rm -rf %s' % self.run_dir).communicate()
--- a/common.py
+++ b/common.py
@@ -82,3 +82,14 @@ def setup_valgrind(mode, name, tmp_dir):
 
     logger.warning('valgrind mode: %s is not supported.', mode)
     return ''
+
+def get_osd_ra():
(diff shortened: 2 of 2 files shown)


roclark:
Hey @markhpc, would this be a sufficient patch for the bug mentioned above? If not, please let me know how I might improve it. Thanks!

mmgaggle:
I like the idea of detecting the read_ahead if nothing is set, this is especially useful with use_existing, where even if osd_ra is set, read_ahead isn't actually set because the set function relies on CBT specific labels. I have a question about get_osd_ra though, how can we ensure that you are getting the read_ahead of the OSD devices, and not the boot disk / journal SSDs, for instance?

roclark:
Very good observation. At present, you are right, we can't ensure that the read_ahead is from the OSDs as opposed to the boot disk or other devices.

My reasoning for the proposed fix is as follows. Based on this code (from cluster/ceph.py)
```python
def set_osd_param(self, param, value):
    common.pdsh(settings.getnodes('osds'), 'find /dev/disk/by-partlabel/osd-device-*data -exec readlink {} \; | cut -d"/" -f 3 | sed "s/[0-9]$//" | xargs -I{} sudo sh -c "echo %s > /sys/block/\'{}\'/queue/%s"' % (value, param))
```
which is called by this block (from benchmark/benchmark.py),
```python
def run(self):
    logger.info('Setting OSD Read Ahead to: %s', self.osd_ra)
    if self.osd_ra:
        self.cluster.set_osd_param('read_ahead_kb', self.osd_ra)
```
aren't we assuming that all of the values for `/sys/block/{}/queue/read_ahead_kb` are the same? At the very end of the line in `set_osd_param`, we echo the input `osd_ra` value to all devices in /sys/block. As such, the value in any of the `/sys/block/{}/queue/read_ahead_kb` should be the same.

Is this a valid assumption? Or is there a more direct way to access the read_ahead value for the OSDs?

mmgaggle:
Yep, that would do the job.

roclark:
Thank you for the feedback! Is there anything else you see that might cause difficulties or you would like me to clarify? Or does this look good for a merge?

mmgaggle:
Yeah, if you make those changes then I think we'll be good to merge. 

@markhpc thoughts?

markhpc:
Looks reasonable to me!

markhpc:
Though in the future please sign the commit!  Thanks. :)

PR #91, opened 2016-02-29
Remove osd_ra requirement for use_existing

mmgaggle:
When someone is using use_existing, we don't set the read ahead
on the OSD devices because we want to leave the cluster "as is".
In this case, we need to set osd_ra to something, so that the
archive directory structure is preserved.


bengland2:
totally agree with this, it's just confusing to CBT users who aren't creating the cluster with CBT.

PR #93, opened 2016-03-23
Added config wizard to easily generate YAML config files

roclark:
This tool is a quick way to generate a YAML config file by stepping the user through the process of specifying values for every CBT setting. Creating a config file manually can be a bit daunting to the first-time user of CBT and this tool aims to solve that issue. In most cases, the user will be given a brief description of the setting and a suggested value. If desired, he or she can also append a '-a' or '--automate' argument to the script to automatically generate a config file without any input, using default values for all settings.

Currently, the tool only supports radosbench, kvmrbdfio and rbdfio, but adding new benchmarks is very straightforward.

### DIFF
--- a/tools/config_class.py
+++ b/tools/config_class.py
@@ -1,0 +1,326 @@
+#!/usr/bin/env python2
+import os
+import sys
+
+MODES = ["randwrite", "randread", "write", "read"]
+
+
+class Config:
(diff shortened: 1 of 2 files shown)


markhpc:
Neat! I haven't really had a chance to look over it in detail yet, but since it's a standalone tool I'll add it in so folks can give it a try.

Thanks!

PR #98, opened 2016-04-20
A few fixes to CBT setup.sh file

ksingh7:
Hello Guys 

With this pull request i am submitting a few fixes to setup.sh, they are :
- Get dependent packages from FC23 repository rather than FC22
- Updated package names for FC23
- FIO installation from source is a 3 step process , configure , make and make install this needs to be done as sudo. 
- updated .gitignore

Changes has been tested and working fine.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -7,7 +7,7 @@ sudo yum check-update
 sudo yum -y update
 sudo yum install -y psmisc util-linux coreutils xfsprogs e2fsprogs findutils \
   git wget bzip2 make automake gcc gcc-c++ kernel-devel perf blktrace lsof \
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -23,9 +23,10 @@ git clone https://github.com/axboe/fio.git
 git clone https://github.com/andikleen/pmu-tools.git
 git clone https://github.com/brendangregg/FlameGraph
 
-cd ${HOME}/fio
-./configure
-make
+cd fio 
+sudo sh configure
+sudo make
+sudo make install
```
ASBishop: The configure and initial make commands should not require sudo. Only the install needs root because the default installation directory is /usr/local.

PR #100, opened 2016-05-11
Add ability to use block size splits

mmgaggle:
Currently CBT can only instruct fio to run workloads that consist
of a single block size. This introduces a way to craft mixed
workloads composed of multiple block sizes by using the bssplit
fio option.

### DIFF
--- a/benchmark/kvmrbdfio.py
+++ b/benchmark/kvmrbdfio.py
@@ -26,7 +26,11 @@ class KvmRbdFio(Benchmark):
         self.rwmixread = config.get('rwmixread', 50)
         self.rwmixwrite = 100 - self.rwmixread
         self.ioengine = config.get('ioengine', 'libaio')
-        self.op_size = config.get('op_size', 4194304)
+        if config.get('bssplit'):
+          self.bssplit = config.get('bssplit')
+          self.op_size = 'bssplit'
+        else:
+          self.op_size = config.get('op_size', 4194304)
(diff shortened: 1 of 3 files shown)


benchmark/kvmrbdfio.py
```diff
@@ -26,7 +26,11 @@ def __init__(self, cluster, config):
         self.rwmixread = config.get('rwmixread', 50)
         self.rwmixwrite = 100 - self.rwmixread
         self.ioengine = config.get('ioengine', 'libaio')
-        self.op_size = config.get('op_size', 4194304)
+        if config.get('bssplit'):
```
ommoreno: This could probably just follow the other parameters' format like:
self.bssplit = config.get('bssplit', None)

benchmark/kvmrbdfio.py
```diff
@@ -93,7 +97,10 @@ def run(self):
         fio_cmd += ' --ramp_time=%s' % self.ramp
         fio_cmd += ' --numjobs=%s' % self.numjobs
         fio_cmd += ' --direct=1'
-        fio_cmd += ' --bs=%dB' % self.op_size
+        if self.bssplit:
```
ommoreno: Similar to above comment, follow other parameters' format:
if self.bssplit is not None:
            fio_cmd += '--bssplit=%s' % self.bssplit
        else:
            fio_cmd += ' --bs=%dB' % self.op_size

PR #104, opened 2016-06-03
rados command should be executed on head node

JevonQ:
The rados command should be executed on head node rather than locally
Signed-off-by: Jevon Qiao <scaleqiao@gmail.com>

### DIFF
--- a/benchmark/radosbench.py
+++ b/benchmark/radosbench.py
@@ -77,8 +77,8 @@ class Radosbench(Benchmark):
 
         if self.concurrent_ops:
             concurrent_ops_str = '--concurrent-ios %s' % self.concurrent_ops
-        #determine rados version    
-        rados_version_str = subprocess.check_output(["rados", "-v"])
+        #determine rados version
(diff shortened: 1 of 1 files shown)


benchmark/radosbench.py
```diff
@@ -77,8 +77,8 @@ def _run(self, mode, run_dir, out_dir):
 
         if self.concurrent_ops:
             concurrent_ops_str = '--concurrent-ios %s' % self.concurrent_ops
-        #determine rados version    
-        rados_version_str = subprocess.check_output(["rados", "-v"])
+        #determine rados version
+        rados_version_str, err = common.pdsh(settings.getnodes('head'), '/usr/bin/rados -v').communicate()
```
ASBishop: I realize a lot of CBT code uses pdsh for sending commands to a single node ('head'), but I find that awkward. Here's an alternative approach that also avoids duplicating a hard-coded command string:

rados_version_str = subprocess.check_output(["ssh", settings.getnodes('head'), self.cluster.rados_cmd, "-v"])

Still, I would not mind seeing this code merged as-is. Hopefully someone with write access to the repo can take a look at this soon.

PR #106, opened 2016-06-10
Execute radosbench test using sudo

### LINKED ISSUE #105: radosbench fails if client cannot read admin keyring
The radosbench test uses the 'rados' tool, which requires the client admin key to access the cluster. The rados command will fail unless the user specified in the config file has read acess to the client admin keyring.

A workaround exists, and that is to set "cmd_path: 'sudo /usr/bin/rados'" in the test configuration. But this is cumbersome and breaks the ability to run the command under valgrind.

ASBishop:
When running a radosbench test, execute the /usr/bin/rados command with sudo
in order to ensure the client can read the admin keyring.

Closes #105


ASBishop:
Rebased to latest master.

PR #110, opened 2016-07-19
Storage pool error check

bengland2:
This change tightens up the error checking from the previous PR #107, which had left pdsh commands unchecked by default.   These checks allow CBT test to abort if early setup stage of test fails, otherwise you can be waiting for a long time to find out that your test did nothing.  They should benefit any of the benchmarks that create storage pools.  The checks are all in mkpool(), which creates the storage pool, and rmpool(), which removes a storage pool.  BTW if a pool doesn't exist then ceph osd pool rm command will not return a failure status, so removal is idempotent.  

Because of the first commit, I discovered a bug that is fixed in the 2nd commit.


markhpc:
I think this looks good.  Have you run it through any cluster creation tests yet?

bengland2:
I just built a cluster using CBT and this PR, but I ran into some problems having nothing to do with my change, I think.    I'll post those as separate issues or PRs but I did get it to run yesterday with this PR and build the cluster.

PR #111, opened 2016-07-25
Kvmrbdfio - support more parameters for wider scope

bengland2:
This change is being proposed to make kvmrbdfio benchmark work better at large scale and also makes it easier to try in small environments.  It includes an example yaml file.  The biggest change is user-specifiable block device list, which means it no longer requires use of separate KVM guests.


mmgaggle:
Probably suitable for a separate PR, but is there any reason we are continuing to use ext4 inside with kvmrbdfio and for rbdfio? We should probably make both benchmarks accept a list of filesystems to iterate through, and default to XFS.

Thoughts?

bengland2:
+1

----- Original Message -----

PR #112, opened 2016-07-25
increase fan-out to support scale-out configurations

bengland2:
These changes help CBT work correctly in scale-out configurations with > 32 client hosts.  This is particularly relevant for the kvmrbdfio benchmark, where the client is a virtual machine and there could be a lot of them.  At present, the default fan-out for pdsh is 32 concurrent sessions, so in this case pdsh will run fio workload on at most 32 hosts at a time.  This change will calculate fan-out required of pdsh to run the workload on all hosts simultaneously.  It handles the "^" syntax to the -w option that indicates the following pathname contains a list of hosts to run the command on, so that CBT log messages aren't mostly a repetitive list of client hosts.

### DIFF
--- a/common.py
+++ b/common.py
@@ -43,16 +43,35 @@ class CheckedPopen:
 # by default, do NOT abort if pdsh returns error status
 # this policy results in minimal code change to CBT while allowing
 # us to strengthen error checking where it's needed.
+#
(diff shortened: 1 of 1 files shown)


markhpc:
This was committed via 2e1fbcd.

PR #114, opened 2016-08-05
Configurable monitoring

bengland2:
This change removes hardcoded perf monitoring tools from CBT, replacing with a pluggable perf monitoring tool infrastructure such that the user can add/remove perf. monitoring as needed.  At the heart of this change is monitoring.py, which now supplies a base class that defines a start(), stop(), and postprocess() method for every perf monitoring sub-class.    The 3 tools that were hardcoded before now live in separate subclass modules.  for perf tool, there is monitor_perf.py, for collectl tool there is monitor_collectl.py, and for blktrace there is monitor_blktrace.py.  The monitoring_factory.py is what instantiates these classes based on the list of comma-separated monitoring tools requested by the user in the monitor_list YAML variable within the benchmark YAML section.  

So far, only one benchmark, librbdfio, has been changed over to use this, as an example.  If there is interest, the remaining benchmarks can be changed over to use this as well.  I tested this on a single VM configured with 3 Ceph OSDs and all the monitoring tools present and running.

### DIFF
--- a/benchmark/librbdfio.py
+++ b/benchmark/librbdfio.py
@@ -65,7 +65,7 @@ class LibrbdFio(Benchmark):
         logger.info('Attempting to populating fio files...')
         if (self.use_existing_volumes == False):
           for i in xrange(self.volumes_per_client):
-              pre_cmd = 'sudo %s --ioengine=rbd --clientname=admin --pool=%s --rbdname=cbt-librbdfio-`hostname -s`-%d --invalidate=0  --rw=write --numjobs=%s --bs=4M --size %dM %s > /dev/null' % (self.cmd_path, self.poolname, i, self.numjobs, self.vol_size, self.names)
+              pre_cmd = 'sudo sh -c "ulimit -n 131072; %s --ioengine=rbd --clientname=admin --pool=%s --rbdname=cbt-librbdfio-`hostname -s`-%d --invalidate=0  --rw=write --numjobs=%s --bs=4M --size %dM %s > /dev/null"' % (self.cmd_path, self.poolname, i, self.numjobs, self.vol_size, self.names)
               p = common.pdsh(settings.getnodes('clients'), pre_cmd,
                               continue_if_error=False)
               ps.append(p)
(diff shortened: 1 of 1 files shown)


markhpc:
Looks fantastic!

bengland2:
Here's an example of what CBT archive looks like with this change.  Note that each monitoring sub-class gets its own subdirectory, this might be different than before.

https://s3.amazonaws.com/ben.england/cbt-sample-archive.tgz

ommoreno:
Hi Ben,

Could you share the yaml file you used for the above sample archive? BTW this looks great!

Thanks,
Orlando

bengland2:
Orlando, so far nothing really changes, except that there is a line in the yaml in the cluster section and in the benchmark section like this:

monitor_list: 'collectl, blktrace, perf'

you can define any subset of these 3 at present, and it's now possible to add new collection tools in a straightforward way.  Mark suggested that each monitoring tool ought to have a section for its configuration (for example, block devices for blktrace).

I might change the name to "monitoring_list" because of the unfortunate name collision between ceph monitors and performance monitoring.  

Glad you like it because it takes a while to test all the benchmarks and cluster configurations, I didn't elect to do all that work until I was sure it was going in the right direction.  So it will take a little time before it's ready to go in to master.

bengland2:
I'm attempting to factor out common monitoring code from benchmark/*.py to either benchmark/benchmark.py or monitoring.py to make it easier to do this for all benchmark/*.py.  This means monitoring would be defined at the cluster level instead of the individual benchmark level, because cluster/ceph.py has no access to individual benchmarks.   any objections on this?  So if you ran a YAML with multiple benchmark sections, this might impact you.  But otherwise I don't think it matters.

bengland2:
I've given up on this one, it's just too big, and I couldn't be sure it wouldn't break something.   If I were to re-attempt it, it would be as a series of much smaller PRs that are more independent of one another, and would be more in collaboration with Mark and others.  Sorry about that.

PR #115, opened 2016-09-01
bugfix of mkpartmagna script

JohnHaan:
- syntax bug of if line


JohnHaan:
The mkpartmagna file has a subtle bug so it doesn't wokr properly.
I input space between $SYSPART and "mapper"

PR #118, opened 2016-09-27
Allow OSD nodes to quiesce after a deleting pool

### LINKED ISSUE #117: Resources required to delete a pool affect subsequent test run
Many benchmarks delete and recreate the test pool between runs. However, the Ceph command to delete a pool returns immediately, and the work of deleting objects in the pool takes place in the background. Unfortunately, experience has shown that disk and CPU resources used while deleting the objects is great enough that it influences the test results for the subsequent run.

One way to avoid the problem is to have the cluster.rmpool() function wait until the disk and CPU utilization on the OSD nodes drops to a reasonable level before returning to the caller. I will be issuing a pull request with this change.

ASBishop:
When deleting a pool, it may take a while for the OSD nodes to delete the
objects in the pool. This change makes CBT wait until the OSD nodes quiesce
in order to ensure they are idle before starting the next test run.

Quiescing is done by waiting until the maximum disk utilization for any
disk falls below 3% across a 30 second window, and waiting until the maximum
CPU utilization for any ceph-osd process falls below 3%.

Closes #117

### DIFF
--- a/cluster/ceph.py
+++ b/cluster/ceph.py
@@ -562,6 +562,17 @@ class Ceph(Cluster):
         common.pdsh(settings.getnodes('head'), 'sudo %s -c %s osd pool delete %s %s --yes-i-really-really-mean-it' % (self.ceph_cmd, self.tmp_conf, name, name),
                     continue_if_error=False).communicate()
 
+        logger.info('Waiting for OSD disk utilization to settle...')
+        disk_util_max = 3
+        window_size = 30
+        wait_cmd = 'while [ $(iostat -dxyz ALL %s 1 | awk \'BEGIN {m=0} {v=int($NF); if(v>m){m=v}} END {print m}\') -gt %s ]; do true; done' % (window_size, disk_util_max)
+        common.pdsh(settings.getnodes('osds'), wait_cmd).communicate()
+
+        logger.info('Waiting for OSD CPU utilization to settle...')
+        osd_cpu_max = 3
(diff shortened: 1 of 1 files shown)


ASBishop:
I'm also using RHEL7.2. I suspect the reason you saw no output is due to the 'z' flag included in the iostat command, so it will produce no output if you happen to run it when disk activity is totally idle. The awk code that processes the output knows how to deal with that situation.

Regarding the CPU usage, I hadn't considered why you might actually want to run CBT while the cluster is scrubbing or in recovery. But the possibility makes sense now that I think about it. But, as you noted, CPU usage will still spike when deleting objects, so I don't know how you distinguish acceptable CPU usage (scrubbing) from the activity we're trying to quiesce (pool deletion).

One idea is to make the thresholds configurable, and include the ability to bypass the quiesce operations entirely. I could add support for the following cluster settings:

```
cluster:
  quiesce_disk_util_max: 3
  quiesce_disk_window_size: 30
  quiesce_osd_cpu_max: 3
```

The settings would be optional, and values I listed would be the defaults. If either "max" setting is < 0 or > 100 then the corresponding quiesce operation would be skipped.

How does that sound?

ASBishop:
I just pushed a new version that makes the quiesce parameters tunable in the config file. The config items are optional, and default to the values I listed above.

mmgaggle:
I tested this, and it's awesome. I was having wild variance between test iterations when the first iteration would create a ton of objects. If I run the same test with this patch, the variance dropped from over 20-100% to less than 5%.

PR #126, opened 2017-04-22
Create raw disk test, useful to get baseline before install ceph

marcosmamorim:


markhpc:
cool, thanks for the new benchmark!

PR #129, opened 2017-09-06
cluster/ceph.py: provide per-pool application information.

markhpc:


markhpc:
neha tested this and it appears to work in her test case.

PR #132, opened 2017-10-04
benchmark: remove directory hierarchy structure of CBT results

neha-ojha:
This PR is aimed at eliminating the directory hierarchy approach of cbt to archive results.
Currently, we have an implementation for the following benchmarks:

- radosbench

- librbdfio
 
We also store the parameters of the cbt run in yaml format inside the archive directory.

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -61,7 +61,7 @@ class Benchmark(object):
         self.cmd_path_full += self.cmd_path
 
         # Store the parameters of the test run
-        config_file = os.path.join(self.archive_dir, 'cbt_config.yaml')
+        config_file = os.path.join(self.archive_dir, 'benchmark_config.yaml')
         if not os.path.exists(self.archive_dir):
             os.makedirs(self.archive_dir)
         if not os.path.exists(config_file):
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -4,6 +4,9 @@ import logging
 import settings
 import common
 import monitoring
+import hashlib
+import os
+import yaml
 
 logger = logging.getLogger('cbt')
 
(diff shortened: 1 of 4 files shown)


markhpc:
I just realized another problem.  Right now CBT skips tests that previously ran if it sees the same directory.  In that way, if you batch up a number of tests you can restart where you left off.  If we go with this new scheme that won't work.  We might want to consider a hash of the benchmark configuration dict instead of a UUID.

neha-ojha:
@markhpc I have addressed (1) and (2). The format is `/tmp/cbt/results/<iteration>/<HASHID>/`
I have also added `/tmp/cbt/results/<iteration>/<HASHID>/cbt_config.yaml` as metadata for each test run. The complete configuration is stored as `/tmp/cbt/results/cbt_config.yaml`

markhpc:
Hi Neha, I tested this and it's working well!

A couple of quick thoughts before merging:  

1) The cbt yaml file is stored as "cbt_config.yaml" in the results directory, but the benchmark configuration yaml is also called "cbt_config.yaml" in the results/<iteration>/<benchmark hash> directory.  I think we should rename the benchmark configuration yaml to something like "benchmark_config.yaml" instead so we don't accidentally get these mixed up.

2) The iteration directory is redundant now since we store the iteration in the benchmark config metadata.  If you want to keep it for teuthology that's fine, but I don't need it anymore on the cbt side.

Other than that, it looks ready for merge!

Mark

neha-ojha:
Hi Mark,

1. Your are right, I have renamed the benchmark config file.
2. For now, I need the iteration dir for cbt-teuthology integration. It helps to easily compare the baseline parameter values against all iterations. It would be great if we could merge it now and if at some point cbt-teuthology does not require it, I will remove it.

Hope I've addressed everything.

PR #135, opened 2017-10-24
add cbt result analysis tool

neha-ojha:
This PR adds a result analysis-tool to the CBT repository. It's job is to compare the test results against baseline parameter values. For now, we have restricted this tool to:

- collect basic parameters like bandwidth, latency, iops etc. for radosbench and librbdfio
- support a single value for each CBT config parameter rather than lists.

The primary purpose of this tool is to compare test results collected by running the CBT task on teuthology against baseline parameter values passed with the CBT task. It marks the job passed/failed based on the baseline and threshold provided. 

One can also choose to run this tool with standalone CBT by adding a "baseline" section to the CBT yaml test configuration. This section should contain baseline numbers for the parameters being collected by the tool and a threshold value.

To use the tool, run the following command inside your CBT repository:
`./analysis-tool.py <cbt_config.yaml> <cbt_results_dir>`

For a librbdfio write test, the baseline section would look something like this:
```
baseline:
...
    threshold: 
    bw: 
    avg_iops: 
    std_dev_iops: 
    avg_lat: 
    std_dev_lat: 
...
```

### DIFF
--- a/analysis-tool.py
+++ b/analysis-tool.py
@@ -1,0 +1,175 @@
+#!/usr/bin/python
+import logging
+import pprint
+import os
+import sys
+import yaml
+import json
+
+from log_support import setup_loggers
+
+logger = logging.getLogger("cbt")
+
+def generate_result_files(dir_path, clients, instances, mode, btype):
+    result_files = []
+    for fname in os.listdir(dir_path):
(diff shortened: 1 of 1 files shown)


neha-ojha:
@jdurgin @markhpc: This is ready for review.

PR #142, opened 2018-02-11
Some minor bug fixes to make librbdfio and radosbench work for Luminous

Jeegn-Chen:


Jeegn-Chen:
@markhpc Could take a look at my fixes?

PR #145, opened 2018-04-02
monitoring.py: use killall instead of pkill

neha-ojha:


onceupon:
may i ask why killall is used here? There are no f flag in killall, i run it with pkill and it works fine.

neha-ojha:
We ran into an issue with pkill sometime back, where it failed to kill the collectl process.

PR #152, opened 2018-09-14
skip sync of archive files on all nodes, the cos log directory will n…

vasukulkarni:
skip sync of archive files on all nodes, the cos log directory will not be available

on all nodes.

Signed-off-by: Vasu Kulkarni <vasu@redhat.com>


vasukulkarni:
@neha-ojha i have tested this with other PR linked here, you can check the logs use different repo for cbt with this change. I will cleanup that PR a bit and push.

PR #156, opened 2019-01-14
Updated parameter in RBD create command

sidhant-agrawal:
- updated librbdfio.py
- updated rbdfio.py
- updated ceph.py
New parameter '--object-size' will be used instead of deprecated parameter '--order' for RBD image creation

Signed-off-by: Sidhant Agrawal <sagrawal@redhat.com>


vasukulkarni:
LGTM

vasukulkarni:
@dillaman any inputs?

vasukulkarni:
@dillaman I dont think we are running anything on jewel yet, I will double check on that, we should be fine with luminous+ runs we are doing now.

vasukulkarni:
I dont see any CBT tests that need backward compatibility, I have also created a branch 'jewel' just in case anything needs to use it for compatibility purpose.

PR #159, opened 2019-02-14
Use iodepth=32 when first writing an RBD image

mogeb:
Signed-off-by: Mohamad Gebai <mgebai@suse.com>


mogeb:
Unfortunately I don't recall the context of this change, and since there doesn't seem to be much value for it, I'll just close this PR. Thanks @markhpc !

PR #160, opened 2019-02-21
client_endpoints: Create new client endpoints and demo fio benchmark.

markhpc:

### DIFF
--- a/client_endpoints/cephfsfuse_client_endpoints.py
+++ b/client_endpoints/cephfsfuse_client_endpoints.py
@@ -1,0 +1,14 @@
+import common
+
+from ceph_client_endpoints import CephClientEndpoints
+
+class CephfsFuseClientEndpoints(CephClientEndpoints):
+    def create(self):
+        self.create_fs()
+
+    def mount(self):
+        self.mount_fs()
+
+    def mount_fs_helper(self, node, dir_name):
+        cmd = 'sudo %s --client_mds_namespace=%s -m %s %s' % (self.ceph_fuse_cmd, self.name, ','.join(self.mon_addrs), dir_name)
+        common.pdsh(node, cmd, continue_if_error=False).communicate()
--- a/client_endpoints/cephfskernel_client_endpoints.py
+++ b/client_endpoints/cephfskernel_client_endpoints.py
@@ -1,0 +1,15 @@
+import common
+
+from ceph_client_endpoints import CephClientEndpoints
+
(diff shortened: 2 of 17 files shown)


markhpc:
- Example cbt yaml file: https://paste.fedoraproject.org/paste/a1y9wHFo58MNs49v1g3qZw
          - Initial test results: https://docs.google.com/spreadsheets/d/1oJZ036QDbJQgv2gXts1oKKhMOKXrOI2XLTkvlsl9bUs/edit?usp=sharing

client_endpoints/cephfsfuse_client_endpoints.py
```diff
@@ -0,0 +1,14 @@
...
+
+class CephfsFuseClientEndpoints(CephClientEndpoints):
+    def create(self):
+        self.create_fs()
+
+    def mount(self):
+        self.mount_fs()
+
+    def mount_fs_helper(self, node, dir_name):
+        cmd = 'sudo mount -t ceph -o mds_namespace=%s %s %s' % (self.name, ','.join(self.mon_addrs), dir_name)
```
batrick: Is this supposed to mount with `ceph-fuse`?

client_endpoints/cephfskernel_client_endpoints.py
```diff
@@ -0,0 +1,14 @@
...
+
+class CephfsKernelClientEndpoints(CephClientEndpoints):
+    def create(self):
+        self.create_fs()
+
+    def mount(self):
+        self.mount_fs()
+
+    def mount_fs_helper(self, node, dir_name):
+        cmd = 'sudo mount -t ceph -o mds_namespace=%s %s %s' % (self.name, ','.join(self.mon_addrs), dir_name)
```
batrick: @ukernel any other options for performance testing?

client_endpoints/cephfsfuse_client_endpoints.py
markhpc: Ugh!  Apparently when I was moving stuff around I must have overwritten it with the kernel version.  I'm testing a fix now.  That also probably explains why the results here look so similar to the kernel version:

https://docs.google.com/spreadsheets/d/1oJZ036QDbJQgv2gXts1oKKhMOKXrOI2XLTkvlsl9bUs/edit?usp=sharing

I'll need to update those.

markhpc:
Added the example as a separate commit, merging this now.

PR #163, opened 2019-03-06
Allow disabling granular fio logs for all backends

mogeb:
Signed-off-by: Mohamad Gebai <mgebai@suse.com>


mogeb:
@neha-ojha this PR is the continuation of #163 , which only allowed `librbdfio` to disable granular logging.

PR #166, opened 2019-06-05
benchmarkfactory.py: remove CephFSFio dependency

neha-ojha:
@markhpc I think this can resolve the issue seen in https://tracker.ceph.com/issues/40120. Let me know if this is what you were thinking.


markhpc:
Yep, that looks right to me!

PR #167, opened 2019-06-06
settings: Don't shutdown when there are no client_endpoints.

markhpc:


neha-ojha:
This makes radosbench pass, librbdfio and cosbench are still failing: http://pulpito.ceph.com/nojha-2019-06-06_21:45:29-perf-basic-master-distro-basic-smithi/

PR #168, opened 2019-06-11
cluster/ceph.py: bring back mkimage()

neha-ojha:


neha-ojha:
This is required to make the client_endpoint changes backwards compatible - to work with existing cbt configuration yaml files.

Before: http://pulpito.ceph.com/nojha-2019-06-06_23:42:35-perf-basic-master-distro-basic-smithi/
After: http://pulpito.ceph.com/nojha-2019-06-11_01:39:36-perf-basic-master-distro-basic-smithi/

markhpc:
Yep, this is necessary until/unless we ditch the librbdfio benchmark completely in favor of the new fio + client_endpoints.  Thanks Neha!

PR #170, opened 2019-06-12
cluster/ceph: Allow string/list representation for mons/osds/clients

markhpc:
This PR allows string and list representation of mons in the yaml to work when building a cluster and using client endpoints.

String representation: 
```yaml
mons: "localhost"
```
or
```yaml
mons: "127.0.0.1"
```
List representation:

```yaml
mons: ["localhost"]
```
or
```yaml
mons:["127.0.0.1"]
```
Dict representation:
```yaml
mons:
  localhost:
    a: "127.0.0.1:6789"
```

usernames for ssh/pdsh can also now be specified on a per-host basis that override the global user setting:

```yaml
mons: "user@localhost"
```
```yaml
mons: ["user@localhost"]
```
```yaml
mons:
  user@localhost:
    a: "127.0.0.1:6789"
```

Currently the string and list representation will assume that mon ids are listed starting with "a" and ending with "z".  dns resolution is used to find the ip to use when creating the monitor which should match what is configured in ceph (via the ceph.conf file or other means).  In the string and list representation it assumed that a single monitor is running on given host on the default port.  If the monitor is running on an alternate ip address, port, uses a different id convention, or multiple monitors are running on a single host, the dict representation should be used.

### DIFF
--- a/cluster/ceph.py
+++ b/cluster/ceph.py
@@ -273,6 +273,42 @@ class Ceph(Cluster):
         common.pdsh(nodes, 'sudo mv /etc/ceph/ceph.conf /etc/ceph/ceph.conf.cbt.bak').communicate()
         common.pdsh(nodes, 'sudo ln -s %s /etc/ceph/ceph.conf' % self.tmp_conf).communicate()
 
+    def get_mon_hosts(self):
+        # get the list of mons
+        mon_hosts  = {}
+        mons_config = settings.cluster.get('mons')
+
+        # For mons specified using the string/list representation, we'll
+        # assume mon ids range from a-z in order and default ports are used.
+        if isinstance(mons_config, str):
+            host = mons_config
+            info = settings.host_info(host)
+            mon_str = '%s:6789' % info['addr']
+            mon_hosts[host] = {'a': mon_str}
+        elif isinstance(mons_config, list):
+            mon_id = 'a'
+            for host in mons_config:
+                info = settings.host_info(host)
+                mon_str = '%s:6789' % info['addr']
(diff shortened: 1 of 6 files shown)


cluster/ceph.py
```diff
@@ -273,6 +274,38 @@ def distribute_conf(self):
...
+            mon_id = 'a'
+            for mon_host in mons_config:
+                mon_addr = socket.gethostbyname(mon_host)
+                mon_hosts[mon_host] = {mon_id: '%s:6789' % mon_addr}
+                if ord(mon_id) < ord('z'):
+                    mon_id = chr(ord(mon_id) + 1)
+                else:
+                    raise ValueError("CBT does not support 27+ monitors")
+        elif isinstance(mon_hosts, dict):
+            # dict that contains hostnames with a mon id / ip:port pair ala:
```
tchaikov: what does "ala" mean?

markhpc: jargon for "like this".  It's redundant though so I just removed it. :)

neha-ojha:
we'll need to address http://pulpito.ceph.com/nojha-2019-06-13_16:33:14-perf-basic-master-distro-basic-smithi/

settings.py
```diff
@@ -78,14 +91,17 @@ def getnodes(*nodelists):
...
+    unique = [node for node in nodes if node]
+    ret = []
+
+    for host in unique:
+        info = host_info(host)
+        host_str = info['host']
+        if 'user' in info:
+            host_str = "%s@%s" % (info['user'], host_str)
+        ret.append(host_str)
+    print ret
```
neha-ojha: might want to get rid of the print here

PR #178, opened 2019-07-17
common: don't require pdsh nor ssh for local operations.

rzarzynski:
This commit allows to run CBT without these dependencies for a scenario where everything happens on the same, local node..

Signed-off-by: Radoslaw Zarzynski <rzarzyns@redhat.com>

### DIFF
--- a/common.py
+++ b/common.py
@@ -1,27 +1,32 @@
 import errno
 import logging
 import os
+import socket
 import subprocess
 
 import settings
 
 logger = logging.getLogger("cbt")
 
+
+def join_nostr(command):
+    return command if isinstance(command, str) else ' '.join(command)
+
 # this class overrides the communicate() method to check the return code and
 # throw an exception if return code is not OK
-
 class CheckedPopen:
     UNINIT=-720
     OK=0
-    def __init__(self, args, continue_if_error=False):
+    def __init__(self, args, continue_if_error=False, shell=False):
+        logger.debug('CheckedPopen continue_if_error=%s, shell=%s args=%s'
+                     % (str(continue_if_error), str(shell), join_nostr(args)))
         self.args = args[:]
         self.myrtncode = self.UNINIT
         self.continue_if_error = continue_if_error
-        self.popen_obj = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
-        logger.debug('CheckedPopen continue_if_error=%s args=%s'%(str(continue_if_error), ' '.join(args)))
+        self.popen_obj = subprocess.Popen(args, shell=shell, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
 
     def __str__(self):
-       return 'checked_Popen args=%s continue_if_error=%s rtncode=%d'%(str(self.args), str(self.continue_if_error), self.myrtncode)
+       return 'checked_Popen args=%s continue_if_error=%s rtncode=%d'%(join_nostr(self.args), str(self.continue_if_error), self.myrtncode)
 
     # we transparently check return codes for this method now so callers don't have to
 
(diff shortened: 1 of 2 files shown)


common.py
```diff
@@ -63,37 +68,67 @@ def expanded_node_list(nodes):
     #logger.info("full list of hosts: %s" % str(full_node_list))
     return node_list
 
+def is_localnode_only(nodes):
+    localnode = get_local_hostname()
+    for node in  expanded_node_list(nodes):
+        if localnode in (node, node.split('@')[1]):
```
rzarzynski: Maybe just:

```python
if localnode == node.split('@')[-1]:
```

markhpc: Yeah, we probably only want to run in local mode if there is only a single host and it's also localhost.  We'll probably need to check to see if it's in string/list/dict form though.  Can we make use of the settings host_info(host) function?  See: https://github.com/ceph/cbt/pull/170/commits/9bd54ac8e53819ad51aabe29349459dbf171bd27

rzarzynski: I see, `host_info()` lets us to avoid duplicating the `@` split. Sure, will change.

common.py
```diff
@@ -63,37 +68,71 @@ def expanded_node_list(nodes):
     #logger.info("full list of hosts: %s" % str(full_node_list))
     return node_list
 
+def is_localnode_only(nodes):
+    # Similarly to `expanded_node_list(nodes)` we assume the passed nodes
```
rzarzynski: @markhpc: switched to `host_info()` and added this comment in the matter of list/dict/str. It seems to me `nodes` can be only a string.

common.py
```diff
@@ -105,6 +144,12 @@ def get_fqdn_list(nodes):
     print ret
     return ret
 
+def get_local_hostname():
```
alfredodeza: system calls like this one are not necessary because Python can do that. Both ceph-deploy and ceph-medic have this utility that checks for the situation of avoiding SSH if it is a local host:

```python
def needs_ssh(hostname, _socket=None):
    """
    Obtains remote hostname of the socket and cuts off the domain part
    of its FQDN.
    """
    if hostname.lower() in ['localhost', '127.0.0.1', '127.0.1.1']:
        return False
    _socket = _socket or socket
    fqdn = _socket.getfqdn()
    if hostname == fqdn:
        return False
    local_hostname = _socket.gethostname()
    local_short_hostname = local_hostname.split('.')[0]
    if local_hostname == hostname or local_short_hostname == hostname:
        return False
    return True
```

rzarzynski: Thanks for pointing this out, will rework.

rzarzynski: Reworked.

common.py
```diff
@@ -1,27 +1,32 @@
...
-    def __init__(self, args, continue_if_error=False):
+    def __init__(self, args, continue_if_error=False, shell=False):
+        logger.debug('CheckedPopen continue_if_error=%s, shell=%s args=%s'
+                     % (str(continue_if_error), str(shell), join_nostr(args)))
         self.args = args[:]
         self.myrtncode = self.UNINIT
         self.continue_if_error = continue_if_error
-        self.popen_obj = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
-        logger.debug('CheckedPopen continue_if_error=%s args=%s'%(str(continue_if_error), ' '.join(args)))
+        self.popen_obj = subprocess.Popen(args, shell=shell, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
```
alfredodeza: Can you explain why is there a need to use `shell=True` (even though the default is False here) ? Seems like it is double the work by having to join the string and then let `Popen` spawn the shell to interpret the command again.

This is potentially buggy, because `shell=True` can have issues with unescaped whitespace, which is not the case then feeding a list of arguments (as it was before)

rzarzynski: > Can you explain why is there a need to use shell=True (even though the default is False here)

I think `ssh` already spawns `/bin/sh` when executing a command. Actually users of `pdsh()` expect it will go through the shell due to e.g. using wildcards when handling file names.
`CheckedPopen` got the `shell` option to avoid duplication. It's set to `True` only by the newly introduced `sh()` that is used instead of `ssh` when executing locally.

markhpc:
So I'm thinking that once everyone is ok with this, let's merge it for now.  Long term I'm thinking we abstract this into a couple of different backend classes (pdsh, cp, native python?) with an abstract interface for remote execution and file movement.  I'm thinking we may want to transition the default away from pdsh since it's getting tougher to get it.  Thoughts?

tchaikov:
@alfredodeza @markhpc gentlemen, is there anything we can do to move proceed?

alfredodeza:
merging for now seems great

PR #179, opened 2019-07-18
 monitoring: integrate perf stat and introduce monitoring_profiles

rzarzynski:


markhpc:
I wonder if we should flesh this out a bit and make profiles for monitoring to run with associated yaml sections.  Something like:

monitoring_profiles:
  perf:
    cmd: "/usr/bin/perf"
   options: ["foo", "bar"]
  collectl:
    cmd: "/usr/bin/collectl"
    options: ["..."]

And then in the benchmark we could do something like:

benchmark:
  fio:
    monitoring_profile: ["perf", "collectl"]
   ...
  ...

What do you think Radek?

rzarzynski:
@markhpc: having this configurable makes a lot of sense – no need to touch the sources to change perf's events, for instance. Will rework.

monitoring.py
```diff
@@ -15,12 +15,14 @@ def start(directory):
 
     # perf
     common.pdsh(nodes, 'mkdir -p -m0755 -- %s' % perf_dir).communicate()
-    # common.pdsh(nodes), 'cd %s;sudo perf_3.6 record -g -f -a -F 100 -o perf.data' % perf_dir)
 
-    osd_nodes = settings.getnodes('osds')
-    common.pdsh(osd_nodes, ['for pid in `cat %s/osd.*.pid`;' % settings.cluster.get('pid_dir'),
-                               'do', 'perf stat -p ${pid} -o %s/perf_stat.${pid} &' % perf_dir,
-                           'done'])
+    perf_nodes = settings.getnodes(settings.cluster.get('monitoring_perf_nodes', 'osds'))
```
rzarzynski: TODO: move to dedicated config section.

monitoring.py
```diff
@@ -1,49 +1,110 @@
...
+        for monitoring, mconfig in sorted(settings.monitoring_profiles.iteritems()):
+            yield Monitoring._get_object(monitoring, mconfig)
+
+    @staticmethod
+    def _get_object(monitoring, mconfig):
+        if monitoring == "collectl":
+            return CollectlMonitoring(mconfig)
+        if monitoring == "perf":
+            return PerfMonitoring(mconfig)
+        if monitoring == "blktrace":
```
rzarzynski: Maybe in the future we could integrate simple (`cmd`, `stop_cmd`, `nodes`) but universal monitoring backend to cover cases without external data dependencies (unlike `perf stat` and its need for knowing pids of OSDs).

markhpc:
This looks great Radek!  I'll try to give it a spin on my test right today.

monitoring.py
```diff
@@ -29,10 +34,12 @@ def stop(directory=None):
 
     common.pdsh(nodes, 'killall -SIGINT -f collectl').communicate()
     common.pdsh(nodes, 'sudo pkill -SIGINT -f perf_3.6').communicate()
+    common.pdsh(nodes, 'sudo pkill -SIGINT -f perf\ stat').communicate()
```
tchaikov: my only concern is that `perf` could be a script which exec the real perf executable. in that case, `pkill` won't be able to kill it

```sh
version="$(uname -r)"
version="${version%%-*}"
case "$version" in
    *.*.*)
    version="${version%.*}"
    ;;
esac
shopt -s execfail
exec "perf_$version" "$@"
```

in my case, `/usr/bin/perf` is something like

```sh
version="$(uname -r)"
version="${version%%-*}"
case "$version" in
    *.*.*)
    version="${version%.*}"
    ;;
esac
shopt -s execfail
exec "perf_$version" "$@"
```

monitoring.py
```diff
@@ -1,56 +1,110 @@
...
+        self.nodes = settings.getnodes(*nodes_list)
 
-def start(directory):
-    nodes = settings.getnodes('clients', 'osds', 'mons', 'rgws')
-    collectl_dir = '%s/collectl' % directory
-    perf_dir = '%s/perf' % directory
-    # blktrace_dir = '%s/blktrace' % directory
+    @staticmethod
+    def _get_all():
+        for monitoring, mconfig in sorted(settings.monitoring_profiles.iteritems()):
```
tchaikov: nit, not py3 compatible. might want to use s/iteritems/items/ ?

monitoring.py
```diff
@@ -1,56 +1,110 @@
...
+    def __init__(self, mconfig):
+        super(BlktraceMonitoring, self).__init__(mconfig)
+        self.osds_per_node = settings.cluster.get('osds_per_node')
+        self.use_existing = settings.cluster.get('use_existing', True)
+        self.user = settings.cluster.get('user')
+
+    def start(self, directory):
+        blktrace_dir = '%s/blktrace' % directory
+        common.pdsh(self.nodes, 'mkdir -p -m0755 -- %s' % blktrace_dir).communicate()
+        for device in xrange (0, self.osds_per_node):
```
tchaikov: nit, not py3 compatible.

could you remove the space after `xrange`?

monitoring.py
```diff
@@ -1,56 +1,110 @@
...
-    perf_dir = '%s/perf' % directory
-    # blktrace_dir = '%s/blktrace' % directory
+    @staticmethod
+    def _get_all():
+        for monitoring, mconfig in sorted(settings.monitoring_profiles.iteritems()):
+            yield Monitoring._get_object(monitoring, mconfig)
+
+    @staticmethod
+    def _get_object(monitoring, mconfig):
+        if monitoring == "collectl":
```
tchaikov: nit, single quote for the sake of consistency.

settings.py
```diff
@@ -25,13 +35,16 @@ def initialize(ctx):
     cluster = config.get('cluster', {})
     client_endpoints = config.get('client_endpoints', {})
     benchmarks = config.get('benchmarks', {})
+    monitoring_profiles = config.get('monitoring_profiles', {})
```
tchaikov: how about moving the backward-compatible hack here? like
```py
monitoring_profiles = config.get('monitoring_profiles', dict(collectl={}))
```
?

rzarzynski: Sure, fixed.

monitoring.py
rzarzynski: Fixed.

monitoring.py
rzarzynski: Fixed and switched to `range`.

monitoring.py
rzarzynski: Switched.

monitoring.py
rzarzynski: Yeah, this is valid point. Need to find a better way. I would prefer to send it as a separated commit to not block this PR.

PR #183, opened 2019-08-06
common: fix local sh execution to return same format as pdsh

markhpc:
https://github.com/ceph/cbt/pull/178/files#diff-cf5a98d3b8ded947cd1a470ce7e3769eR86-R88

introduced a bug where commands executed via sh didn't return the same output as pdsh.  pdsh always prepends the hostname the command was run on with a colon and a space to stdout and stderr while sh does not.  This PR is a bit of a hack that makes it so that local sh execution of commands also prepends the fqdn of the local host to the stderr and stdout output.

Eventually, we probably want to make this cleaner by returning a dict with the stdout, stderr, host, and potentially other information in a more generic way, but for now this gets things working again.

### DIFF
--- a/common.py
+++ b/common.py
@@ -14,7 +14,7 @@ def join_nostr(command):
 
 # this class overrides the communicate() method to check the return code and
 # throw an exception if return code is not OK
-class CheckedPopen:
+class CheckedPopen(object):
     UNINIT=-720
     OK=0
     def __init__(self, args, continue_if_error=False, shell=False):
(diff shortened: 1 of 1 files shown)


common.py
```diff
@@ -45,6 +45,13 @@ def wait(self):
         self.communicate(continue_if_error=True)
         return self.myrtncode
 
+class CheckedPopenLocal(CheckedPopen):
+    def communicate(self, input=None, continue_if_error=True):
+        local_fqdn = get_fqdn_local()
```
markhpc: Turns out this isn't good enough.  We need to prepend whatever the value was that passed the is_localnode_only check.  Since we have to do the prepend in communicate(), we'll likely need to pass the node list in when we create CheckedPopen though this is pretty ugly.

common.py
```diff
@@ -45,6 +45,17 @@ def wait(self):
         self.communicate(continue_if_error=True)
         return self.myrtncode
 
+class CheckedPopenLocal(CheckedPopen):
+    def __init__(self, host, args, continue_if_error=False, shell=False):
+        self.host = host
+        super(CheckedPopenLocal, self).__init__(args, continue_if_error, shell)
```
tchaikov: nit, i'd suggest initialize the super class first.

common.py
```diff
@@ -80,16 +91,17 @@ def is_localnode_only(nodes):
...
-    return CheckedPopen(join_nostr(command),
-                        continue_if_error=continue_if_error, shell=True)
+def sh(local_node, command, continue_if_error=True):
+    return CheckedPopenLocal(local_node, join_nostr(command),
+                             continue_if_error=continue_if_error, shell=True)
 
 def pdsh(nodes, command, continue_if_error=True):
-    if is_localnode_only(nodes):
-        return sh(command, continue_if_error=continue_if_error)
+    local_node = is_localnode_only(nodes);
```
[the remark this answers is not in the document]
markhpc: done!

common.py
markhpc: done!

common.py
markhpc: fixed!

PR #182, opened 2019-08-06
cluster/ceph: Fix conf file regression.

markhpc:
Signed-off-by: Mark Nelson <mnelson@redhat.com>

PR #184, opened 2019-08-08
settings: better fix for conf and archive dir checks.

markhpc:
https://github.com/ceph/cbt/pull/182 didn't actually fix the conf file regression seen in teuthology since the conf file was being set improperly in settings.  Revert that change, fix settings.py, and perform error checking for the conf_file and archive_dir.

Edit: Also fix a bug where the command line archive_dir setting was always being used for directory creation.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -43,13 +43,19 @@ def initialize(ctx):
     if not benchmarks:
         shutdown('No benchmarks section found in config file, bailing.')
 
+    # set the archive_dir from the commandline if present
+    if ctx.archive:
+        cluster['archive_dir'] = ctx.archive
+    if 'archive_dir' not in cluster:
+        shutdown('No archive dir has been set.')
+
     _handle_monitoring_legacy()
 
     # store cbt configuration in the archive directory
-    cbt_results = os.path.join(ctx.archive, 'results')
+    cbt_results = os.path.join(cluster['archive_dir'], 'results')
     config_file = os.path.join(cbt_results, 'cbt_config.yaml')
-    if not os.path.exists(ctx.archive):
-        os.makedirs(ctx.archive)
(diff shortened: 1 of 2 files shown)


settings.py
```diff
@@ -61,14 +67,19 @@ def initialize(ctx):
...
+    # If no conf file is set, default to /etc/ceph/ceph.conf
+    # FIXME: We shouldn't have cluster specific defaults in settings.
+    # Eventually make a base class with specific cluster implementations.
+    if 'conf_file' not in cluster:
+        cluster['conf_file'] = '/etc/ceph/ceph.conf'
+    try:
+        f = open(cluster['conf_file'])
+        f.close()
+    except IOError, e:
+        shutdown('Was not able to access conf file: %s' % cluster['conf_file'])
```
tchaikov: how about
```py
if not os.access(cluster['conf_file'], os.R_OK):
  shutdown('Was not able to access conf file: %s' % cluster['conf_file'])
```
?

neha-ojha:
@markhpc http://pulpito.ceph.com/nojha-2019-08-09_20:03:58-perf-basic-master-distro-basic-smithi/ - some tests failed due to incorrect hostname parsing

markhpc:
@neha-ojha ok, that looks like a different bug in #178 related to https://github.com/ceph/cbt/pull/178/files#diff-cf5a98d3b8ded947cd1a470ce7e3769eR12-R13.

rzarzynski:
@markhpc, @neha-ojha: apologizes for this bug. It looks the recent changes to `monitoring.py` pass the commands to `pdsh()` as lists of strings while the function expects just strings. Created #186 to fix that.

markhpc:
Closed in favor of bigger fix in #189

PR #187, opened 2019-08-12
monitoring: kill local "perf" processes by their pids

tchaikov:
as on some distro, `perf` actuall execv another `perf_*` to do its job,
so we cannot use `pkill` to kill `perf`

### DIFF
--- a/common.py
+++ b/common.py
@@ -1,6 +1,7 @@
 import errno
 import logging
 import os
+import signal
 import socket
 import subprocess
 
@@ -23,7 +24,12 @@ class CheckedPopen(object):
         self.args = args[:]
         self.myrtncode = self.UNINIT
         self.continue_if_error = continue_if_error
-        self.popen_obj = subprocess.Popen(args, shell=shell, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+        self.shell = shell
+        self.popen_obj = subprocess.Popen(args, shell=shell,
+                                          stdout=subprocess.PIPE,
+                                          stderr=subprocess.PIPE,
+                                          preexec_fn=os.setsid,
+                                          close_fds=True)
 
     def __str__(self):
        return 'checked_Popen args=%s continue_if_error=%s rtncode=%d'%(join_nostr(self.args), str(self.continue_if_error), self.myrtncode)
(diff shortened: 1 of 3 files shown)


monitoring.py
```diff
@@ -47,18 +50,34 @@ def __init__(self, mconfig):
...
+                with file(pid_path) as pidfile:
+                    pid = pidfile.read().strip()
+                    perf_cmd = perf_template.format(perf_dir=perf_dir, pid=pid)
+                    runner = common.sh(local_node, perf_cmd)
+                    self.perf_runners.append(runner)
+        else:
+            # ${pid} will be handled by remote's sh
+            perf_cmd = perf_template.format(perf_dir=perf_dir, pid='${pid}')
+            common.pdsh(self.nodes, ['for pid in `cat %s/osd.*.pid`;' % self.pid_dir,
+                                     'do', perf_cmd,
```
rzarzynski: The remote perf shutdown seems to be painful. Maybe executing the command through a helper (something like `start-stop-daemon`) to catch and store PID into a file is the way to go in long term. Still, proper handling on local is significant step forward :+1:.

tchaikov: if we are going to use start-stop-daemon, probably we should always use it no matter it's local or remote?

tchaikov: another way is to deploy a snippet of the python script to do `start-stop-daemon`?

rzarzynski: > if we are going to use start-stop-daemon, probably we should always use it no matter it's local or remote?

Yup, I think ultimately we will reunify that.

> another way is to deploy a snippet of the python script to do `start-stop-daemon`?

Possibly but this would require some code lifting (scp) between nodes.

common.py
```diff
@@ -23,7 +24,12 @@ def __init__(self, args, continue_if_error=False, shell=False):
         self.args = args[:]
         self.myrtncode = self.UNINIT
         self.continue_if_error = continue_if_error
-        self.popen_obj = subprocess.Popen(args, shell=shell, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+        self.shell = shell
+        self.popen_obj = subprocess.Popen(args, shell=shell,
+                                          stdout=subprocess.PIPE,
+                                          stderr=subprocess.PIPE,
+                                          preexec_fn=os.setsid,
```
rzarzynski: Makes sense.

PR #190, opened 2019-08-15
fio_objectstore_tools: utilities for tracing fio_ceph_objectstore

athanatos:
This adds a few python scripts for initializing, running, and
analyzing benchmark runs using the fio_ceph_objectstore backend.

The tooling is focused on understanding the
latency/throttle/throughput relationship for a particular
device.

Signed-off-by: Samuel Just <sjust@redhat.com>

### DIFF
--- a/tools/fio_objectstore_tools/analyze.py
+++ b/tools/fio_objectstore_tools/analyze.py
@@ -1,0 +1,124 @@
+#!env python3
+
+import json
+import os
+import sys
+import itertools
+import argparse
+import sys
+import subprocess
+import numpy as np
+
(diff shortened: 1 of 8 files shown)


athanatos:
The next step would be to incorporate support for running the lttng sessions and collecting trace results into cbt proper.

athanatos:
@markhpc Does this look appropriate?

markhpc:
@athanatos looks good, though maybe move the fio_objecstore_tools directory under the top level tools directory?

athanatos:
@markhpc will do

PR #191, opened 2019-08-22
cosbench: specify secret-key when creating a key and cleanups

tchaikov:


tchaikov:
http://pulpito.ceph.com/kchai-2019-08-22_12:03:40-perf-basic-master-distro-basic-smithi/4240984

PR #198, opened 2019-09-12
common: do not fail if output is not encoded in default encoding

tchaikov:


tchaikov:
http://pulpito.ceph.com/kchai-2019-09-12_04:33:05-rados:perf-master-distro-basic-smithi/

PR #199, opened 2019-09-26
radosbench: fix version extraction and some behaviors depended on it

tchaikov:


tchaikov:
tested on incerta08. see https://github.com/ceph/ceph/pull/30578#issuecomment-535330001

PR #205, opened 2019-12-04
common: Don't invoke local cmd logic when there are multiple nodes

markhpc:


markhpc:
This PR is to fix a bug where one of the nodes in the nodes list maps to the local machine which ends up overriding the entire nodes list.  Instead if there are multiple nodes in the list, just fall back to pdsh.

PR #207, opened 2019-12-17
Added new metric to Rados Bench - CPU cycles per operation and Fixing the cbt config file generator not including monitoring profiles

AmnonHanuhov:
I made 3 main changes:
1. Added a metric for Rados Bench to be able to present the number of cpu cycles per operation
2. Fixed an issue where setting.py wasn't preparing the cbt config file correctly (Monitoring profiles were not added)
3. Cleaned the name-space in radosbench.py by giving more accurate names to some of the variables

PR #211, opened 2020-03-02
Generalize cycles-per-op to support multiple types of benchmarks

AmnonHanuhov:
I introduced an interface called DataAnalyzer that has metrics that are of interest to us and every benchmark should support. Prior to my change, the implementation of fetching results from the benchmarks was specific to rados bench (e.g assuming the results are in a json file). The idea is to create an analyzer class for each benchmark that will implement the result fetching in it's own way, and then functions like evaluate() and compare_client_results() can be completely independent from the type of benchmark used.


AmnonHanuhov:
@rzarzynski @tchaikov

benchmark/radosbench.py
```diff
@@ -253,18 +253,9 @@ def _compare_client_results(self, client_run, fname_from_evaluate, out_dirs):
...
 
-        
-
-        for fname in fnames:
+        for fname in results_files_paths:
             with open(fname) as f:
                 json_outputs.append(json.load(f))
-                
-        i = 0
-        for fpath in out_dirs:
```
rzarzynski: Hmm, this commit isn't just a rename.

benchmark/radosbench.py
```diff
@@ -280,16 +298,14 @@ def evaluate(self, baseline):
         for run in runs:
             out_dirs = [os.path.join(self.out_dir, run),
                         os.path.join(baseline.out_dir, run)]
-            for d in out_dirs:
```
rzarzynski: The debug from previous commit is removed by this one. I think we should rebase the branch to exclude these changes entirely from the history.

monitoring.py
```diff
@@ -155,11 +154,11 @@ def monitor(directory):
...
-        if(isinstance(monitoring_profile, PerfMonitoring):
-            return monitoring_profile.get_cpu_cycles() #if it is, then return the number of cycles
-    #else
-        #throw an exception or something
+    # check if perf stat is configured
+    for monitoring_profile in Monitoring._get_all():
+        if(isinstance(monitoring_profile, PerfMonitoring)):
+            return monitoring_profile.get_cpu_cycles(out_dir) # if it is, then return the number of cycle
+    raise Exception("Perf monitor not configured!")
+
```
rzarzynski: Hmm, strange. Many changes overlap with the previous commits. Is that an artifice of an unclean rebase?

setup.sh
```diff
@@ -28,8 +27,8 @@ cd ${HOME}/fio
 ./configure
 make
 
-#wget https://access.redhat.com/downloads/content/281/ver=3/rhel---7/3.3/x86_64/product-software
-sudo mount -o loop Ceph-*-dvd.iso /mnt
+# wget < Red Hat Ceph Storage ISO URL >
+# sudo mount -o loop Ceph-*-dvd.iso /mnt
```
rzarzynski: This commit seems to be actually a fixup for https://github.com/ceph/cbt/pull/211/commits/90b600f519a1475f656877db991f300a4749e744. If so, they should be squashed. Though, the rebase to current variant of #207 could fix this up.

benchmark/radosbenchanalyzer.py
```diff
@@ -0,0 +1,33 @@
+from .dataanalyzer import DataAnalyzer
+
+class RadosBenchAnalyzer implements DataAnalyzer
```
rzarzynski: I see this is actually a pre-prototype of actual implementation of `DataAnalyzer`. Still, it shouldn't lead to `SyntaxError`.

benchmark/dataanalyzer.py
```diff
@@ -0,0 +1,23 @@
+from abc import ABC, abstractmethod
+
+class DataAnalyzer(ABC):
+
+    def __init__(self, archive_dir):
+        super.__init__()
+        self.archive_dir = archive_dir
+
+    @abstractmethod
+    def create_data_analyzer():
```
rzarzynski: This method doesn't belong here. It's rather an attribute of `Benchmark` and its particular implementations.

monitoring.py
rzarzynski: The commit is a fixup and should be squashed.

benchmark/radosbenchanalyzer.py
```diff
@@ -1,23 +1,29 @@
 from .dataanalyzer import DataAnalyzer
+import os
+import json
+import monitoring
 
-class RadosBenchAnalyzer implements DataAnalyzer 
-    def __init__(self, archive_dir):
-        super(RadosBenchAnalyzer, self).__init__(archive_dir)
+class RadosBenchAnalyzer(DataAnalyzer):
```
rzarzynski: OK, this seems to solve the `SyntaxError`. Please squash with 49e89f4ca4aaf32a442cb6993c4dde0e6893054f.

benchmark/benchmark.py
```diff
@@ -41,6 +43,70 @@ def __init__(self, archive_dir, cluster, config):
...
+        # normalize the names
+        aliases = {'bandwidth': 'Bandwidth (MB/sec)',
+                   'iops_avg': 'Average IOPS',
+                   'iops_stddev': 'Stddev IOPS',
+                   'latency_avg': 'Average Latency(s)'}
+        res_outputs = [] # list of dictionaries containing the self and baseline benchmark results
+        compare_results = []
+        self_analyzer_res = {}
+        baseline_analyzer_res = {}
+        self_analyzer_res.update([(aliases['bandwidth'], self_analyzer.get_bandwidth()),
```
rzarzynski: Why to `update`? `self_analyzer_res` is defined right above.

benchmark/benchmark.py
```diff
@@ -41,6 +43,70 @@ def __init__(self, archive_dir, cluster, config):
...
+                                  (aliases['iops_avg'], self_analyzer.get_average_iops()),
+                                  (aliases['iops_stddev'], self_analyzer.get_stddev_iops()),
+                                  (aliases['latency_avg'], self_analyzer.get_average_latency())])
+        baseline_analyzer_res.update([(aliases['bandwidth'], baseline_analyzer.get_bandwidth()),
```
rzarzynski: Same here.

benchmark/benchmark.py
```diff
@@ -41,6 +43,70 @@ def __init__(self, archive_dir, cluster, config):
...
+                                  (aliases['iops_avg'], baseline_analyzer.get_average_iops()),
+                                  (aliases['iops_stddev'], baseline_analyzer.get_stddev_iops()),
+                                  (aliases['latency_avg'], baseline_analyzer.get_average_latency())])
+        res_outputs.append(self_analyzer_res)
+        res_outputs.append(baseline_analyzer_res)             
+        self_num_cpu_cycles_per_op = self_analyzer.get_cpu_cycles_per_op()
+        if self_num_cpu_cycles_per_op != None: # perf stat monitor is configured
```
rzarzynski: The comment is about an implementation detail of `Radosbench`. It's good idea to mark this down.

benchmark/benchmark.py
```diff
@@ -41,6 +43,70 @@ def __init__(self, archive_dir, cluster, config):
...
+            res_outputs[0]['Cycles per operation'] = self_num_cpu_cycles_per_op
+            res_outputs[1]['Cycles per operation'] = baseline_analyzer.get_cpu_cycles_per_op()
+            self.acceptable.update({'cpu_cycles_per_operation':'(or (less) (near 0.05))'})
```
rzarzynski: I bet we want to update the YAML instead of appending the criterion to every run.

benchmark/benchmark.py
rzarzynski: Another idea:

```python
for alias, stmt in list(self.acceptable.items()):
    # ...
    getter = getattr(analyzer, 'get_'  + alias)
    value = getter()
    if value == None:
        # the case of e.g. cycles-per-op while no monitoring is configured.
        # msg about ignoring here.
        continue
```

AmnonHanuhov: From my understanding it is better to use update() if you want to add multiple keys at once

benchmark/benchmark.py
AmnonHanuhov: Read comment above

benchmark/radosbench.py
```diff
@@ -244,17 +244,32 @@ def analyze(self, out_dir):
...
         compare_results = []
-        for fname in fnames:
+        for output_number, fname in enumerate(results_files_paths):
             with open(fname) as f:
                 json_outputs.append(json.load(f))
+                total_ops = self.get_total_ops(json_outputs[output_number])
+                num_cpu_cycles = self.get_cpu_cycles(out_dirs[output_number])
+                if num_cpu_cycles != None:
+                    json_outputs[output_number]['Cycles per operation'] = int(num_cpu_cycles) / int(total_ops)
+                    self.acceptable.update({'cpu_cycles_per_operation':'(or (less) (near 0.05))'})
```
tchaikov: i'd avoid encoding the acceptance criteria in CBT. we should put them in the test case.

benchmark/radosbench.py
```diff
@@ -244,17 +244,32 @@ def analyze(self, out_dir):
...
 
-    def _compare_client_results(self, client_run, fnames):
+    def get_total_ops(self, json_output):
+        search_key = "made" # looking for either 'Total writes made' or 'Total reads made'
+        res = [val for key, val in json_output.items() if search_key in key]
+        return res[0]
+
+    # getting the number of cpu cycles calculated by Perf Stat
+    def get_cpu_cycles(self, perf_out_dir):
+        return monitoring.get_cpu_cycles(perf_out_dir)
```
tchaikov: instead of analyzing the perf stats, i'd suggest consolidate this into `Radosbench.analyze()`. and add the parsed result to `json_output*`.

tchaikov:
@AmnonHanuhov ping? did you get a chance to reconsider and address @rzarzynski 's comments?

benchmark/radosbench.py
AmnonHanuhov: Changed this in 6e3e3d2, I added cpu cycles per op as an acceptance criteria in CBT and I made a seperate pull request to change the YAML (here's a link: https://github.com/ceph/ceph/pull/34003)

benchmark/benchmark.py
AmnonHanuhov: Thanks for the advice, I used it to make a similar implementation in 055875c

benchmark/benchmark.py
AmnonHanuhov: True, fixed this in the latest commit, you can see my comment to Kefu down below.

benchmark/benchmark.py
AmnonHanuhov: There was no need for update in the new way I implemented it so I got rid of it.

benchmark/benchmark.py
AmnonHanuhov: Changed the condition and the comment accordingly.

setup.sh
```diff
@@ -11,14 +11,14 @@ sudo yum install -y psmisc util-linux coreutils xfsprogs e2fsprogs findutils \
 
 MIRROR="http://mirror.hmc.edu/fedora/linux/releases/22/Everything/x86_64/os/Packages"
 
-wget ${MIRROR}/p/pdsh-2.31-3.fc22.x86_64.rpm
```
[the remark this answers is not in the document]
rzarzynski: Apart from killing the newlines, the commit is now separated. The title (*Added functionality to be able to present cpu cycles per operation*) doesn't actually fit the patch. However, It looks the commit is just an artifact and should be squashed with the following one.

setup.sh
```diff
@@ -9,16 +9,15 @@ sudo yum install -y psmisc util-linux coreutils xfsprogs e2fsprogs findutils \
   git wget bzip2 make automake gcc gcc-c++ kernel-devel perf blktrace lsof \
   redhat-lsb sysstat screen python3-yaml ipmitool dstat zlib-devel ntp
 
-MIRROR="http://mirror.hmc.edu/fedora/linux/releases/22/Everything/x86_64/os/Packages"
```
rzarzynski: OK, the title reflects the changes. Please squash the commit with previous one.

rzarzynski: Removed.

benchmark/benchmark.py
```diff
@@ -8,6 +8,8 @@
 import os
 import json
 import yaml
+from .radosbenchanalyzer import RadosBenchAnalyzer
+from .lis import Lispy, Env
```
markhpc: Generally I try to keep the number of imports as low as possible for the core CBT modules.  Can we do the same thing without the import?

benchmark/dataanalyzer.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABC, abstractmethod
+
+class DataAnalyzer(ABC):
```
markhpc: Can this class be moved directly into benchmark.py?

benchmark/radosbenchanalyzer.py
```diff
@@ -0,0 +1,42 @@
+from .dataanalyzer import DataAnalyzer
+import os
+import json
+import monitoring
+
+class RadosBenchAnalyzer(DataAnalyzer):
```
markhpc: Like above, any way we can move these Analyzer classes into the various benchmark modules?

benchmark/radosbenchanalyzer.py
```diff
@@ -0,0 +1,42 @@
+from .dataanalyzer import DataAnalyzer
+import os
+import json
+import monitoring
+
+class RadosBenchAnalyzer(DataAnalyzer): 
+    def __init__(self, archive_dir, run, host, proc):
+        super(RadosBenchAnalyzer, self).__init__(archive_dir, run, host, proc)
```
tchaikov:
```suggestion
        super().__init__(archive_dir, run, host, proc)
```

benchmark/radosbenchanalyzer.py
```diff
@@ -0,0 +1,42 @@
...
+        self.out_dir = os.path.join(self.archive_dir, run)
+        self.radosbench_out_fname = os.path.join(self.out_dir, 'json_output.{proc}.{host}'.format(proc=proc,
+                                                               host=host))
```
tchaikov:
```suggestion
        self.radosbench_out_fname = os.path.join(self.out_dir, f'json_output.{proc}.{host}')
```

benchmark/radosbench.py
tchaikov: @AmnonHanuhov could you please remove this line in this very commit. and add the alias in this commit?

benchmark/radosbench.py
```diff
@@ -244,17 +244,32 @@ def analyze(self, out_dir):
...
-        for fname in fnames:
+        for output_number, fname in enumerate(results_files_paths):
             with open(fname) as f:
                 json_outputs.append(json.load(f))
+                total_ops = self.get_total_ops(json_outputs[output_number])
+                num_cpu_cycles = self.get_cpu_cycles(out_dirs[output_number])
+                if num_cpu_cycles != None:
+                    json_outputs[output_number]['Cycles per operation'] = int(num_cpu_cycles) / int(total_ops)
+                    self.acceptable.update({'cpu_cycles_per_operation':'(or (less) (near 0.05))'})
+                    aliases.update({'cpu_cycles_per_operation': 'Cycles per operation'})
```
tchaikov: again, please squash related changes into a single one.

benchmark/benchmark.py
tchaikov: @AmnonHanuhov could you note down the reason why we need to

> Moved compare_client_results() and evaluate() from radosbench.py to benchmark.py

?

benchmark/radosbenchanalyzer.py
AmnonHanuhov: Ok I made the suggested change

benchmark/radosbenchanalyzer.py
AmnonHanuhov: Made this suggested change too

compare.py
```diff
@@ -67,3 +67,4 @@ def main():
 
 if __name__ == '__main__':
     main()
+
```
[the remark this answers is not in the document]
AmnonHanuhov: Removed the unnecessary lines

monitoring.py
```diff
@@ -1,10 +1,11 @@
 from contextlib import contextmanager
 from glob import glob
 import os.path
-
 import common
 import settings
 
+
```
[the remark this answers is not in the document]
AmnonHanuhov: Thanks for the reminder, I forgot about it. I got rid of the extra new lines.

benchmark/radosbench.py
AmnonHanuhov: I'm not entirely sure which changes you wanted me to squash, can you elaborate please? It seems like all the changes to radosbench.py are indeed under one commit apart from the commit where I moved evaluate and compare_client_results to benchmark. The commit you are commenting on was probably carried over from the previous PR I made, so do you want me to just get rid of it as these functions are in a different file now anyway?

benchmark/benchmark.py
```diff
@@ -8,6 +8,8 @@
 import os
 import json
 import yaml
+from .radosbenchanalyzer import RadosBenchAnalyzer
```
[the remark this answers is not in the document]
AmnonHanuhov: This was a mistake, it actually should have been DataAnalyzer. I fixed it.

benchmark/benchmark.py
AmnonHanuhov: Well the Lispy import is used in compare_client_results() for parsing the results. This is actually nor an import I introduced nor was the use of Lispy (Take a look at the current compare_client_results() code in radosbench.py in cbt/master). I moved this part of code from radosbench.py. It seems like it is necessary here unless we can find another way to parse out the results.

benchmark/dataanalyzer.py
AmnonHanuhov: I think it is possible, meaning it won't break the code and it'll have the exact same behavior, but I'm not entirely sure if in terms of code design it should be part of benchmark.py. In my opinion the separation is good but I have no problem changing it if that's what you want. Let me know what you think.

AmnonHanuhov: Ok so eventually I decided to move these classes inside the various benchmarks like you asked and removed the extra files. You can see the changes in the latest commit.

benchmark/benchmark.py
AmnonHanuhov: To overcome this issue while still using Lispy, I just moved the import inside the compare_client_results() function (which is the only function using it in benchmark.py).

AmnonHanuhov: @tchaikov Compare and Evaluate are functions that are not specific to radosbench, if you take a look at their implementation, they are completely independent from radosbench specific settings. The purpose of this PR is to generalize CBT and introduce a new metric, for this reason, it is only appropriate that these functions will be moved so that other benchmarks like fio, rbdfio, radosbench etc

benchmark/radosbench.py
AmnonHanuhov: Ok, I think I squashed the commits you were talking about. Check it out and let me know

markhpc:
LGTM

PR #220, opened 2020-06-11
cluster/ceph: do not specify asok path explicitly

tchaikov:
there is chance that we run cbt against a vstart cluster which puts the
asok files in $TMPDIR instead of /var/run/ceph. so. in this chance,
use `ceph daemon` instead of `ceph --admin-daemon <path>` to let `ceph`
cli to deduce the path of asock by itself to avoid hardwire the path to
/var/run.

### DIFF
--- a/cluster/ceph.py
+++ b/cluster/ceph.py
@@ -608,7 +608,7 @@ class Ceph(Cluster):
             time.sleep(1)
 
     def dump_config(self, run_dir):
-        common.pdsh(settings.getnodes('osds'), 'sudo %s -c %s --admin-daemon /var/run/ceph/ceph-osd.0.asok config show > %s/ceph_settings.out' % (self.ceph_cmd, self.tmp_conf, run_dir)).communicate()
+        common.pdsh(settings.getnodes('osds'), 'sudo %s -c %s daemon osd.0 config show > %s/ceph_settings.out' % (self.ceph_cmd, self.tmp_conf, run_dir)).communicate()
 
     def dump_historic_ops(self, run_dir):
         common.pdsh(settings.getnodes('osds'), 'find "/var/run/ceph/ceph-osd*.asok" -maxdepth 1 -exec sudo %s --admin-daemon {} dump_historic_ops \; > %s/historic_ops.out' % (self.ceph_cmd, run_dir)).communicate()


tchaikov:
@markhpc ceph cli assumes `ceph-conf` is listed in `$PATH`. and it also sets `$PATH` in developer mode, so it works with a vstart cluster.

i think you 

1. installs ceph into `/usr/local` using `make install`
1. does not have `/usr/local` in your `$PATH`

this is unusual. so i don't think it's an issue bothering ceph cli. actually, it's an issue bothering you.

markhpc:
@tchaikov It is standard for distributions to not have /usr/local in root's path to avoid accidentally running external commands with root privileges.  This is almost universal across distributions afaik.

Given your description, does this sound right?

1) The ceph cli requires ceph-conf to be in the path when using the daemon command
2) The ceph cli must be run with superuser privileges (permission denied due to default /var/run/ceph permissions) when issuing admin socket commands.
3) ceph from source installs in /usr/local by default
4) distributions don't put /usr/local/bin in path by default
5) the combination of 1-4 mean that by default ceph daemon commands break with source builds without workarounds:

a) put /usr/local/lib in root/sudo path (bad!)
b) install ceph somewhere in root/sudo path (also bad if you are trying to keep root's path pure)
c) make /var/run/ceph have 777 perms (not great, but what teuthology apparently does!):
  https://tracker.ceph.com/issues/5972
d) change admin socket directory in ceph.conf and run ceph cli as non-root user (better than a/b/c, but still broken for root)
e) use ceph cli with the --admin-daemon flag instead of daemon to avoid the issue (what cbt currently does)

I would be far more OK with this PR if I wasn't worried about the ceph cli breaking behind cbt's back with unhelpful error messages that can confuse the user.  Is there something I'm missing?

tchaikov:
> @tchaikov It is standard for distributions to not have /usr/local in root's path to avoid accidentally running external commands with root privileges. This is almost universal across distributions afaik.
> 
> Given your description, does this sound right?
> 
> * The ceph cli requires ceph-conf to be in the path when using the daemon command
> * The ceph cli must be run with superuser privileges (permission denied due to default /var/run/ceph permissions) when issuing admin socket commands.
> * ceph from source installs in /usr/local by default
> * distributions don't put /usr/local/bin in the root/sudo path by default

i think a user which uses "make install" to install the artifacts into `/usr/local` should rely on the distro's default settings.  neither does `pip install --user` install the python package into system path, user is on his/her own to set the `$PATH`.

> 
> The combination of these mean that by default ceph daemon commands break with source builds without workarounds afaik:
> 
> 1. put /usr/local/lib in root/sudo path (bad!)
> 2. install ceph somewhere in root/sudo path (also bad if you are trying to keep root's path pure and /usr/bin pure)
> 3. make /var/run/ceph have 777 perms (not great, but what teuthology apparently does!):
>    https://tracker.ceph.com/issues/5972
> 4. change admin socket directory in ceph.conf and run ceph cli as non-root user (better than 1-3, but still broken for root)
> 5. use ceph cli with the --admin-daemon flag instead of daemon to avoid the issue (what cbt currently does)
> 
> I would be far more OK with this PR if I wasn't worried about the ceph cli breaking behind cbt's back with unhelpful error messages that can confuse the user. Is there something I'm missing?

i don't think ceph's install script should touch user's `$PATH`.

tchaikov:
i went ahead and fixed `ceph.in`, see https://github.com/ceph/ceph/pull/35592

tchaikov:
@markhpc i think i've fixed the your use case, could you take another look?

PR #222, opened 2020-07-12
compare, tools: add github check support

tchaikov:


compare.py
```diff
@@ -50,20 +89,38 @@ def main():
...
+        else:
+            heading = Heading3(f'all {nr_tests} tests passed')
+        ctx.output.write(str(heading))
+
+        table = Table()
+        table.add_headers('run', 'metric', 'baseline', 'result', 'accepted')
+        for r in results:
+            table.add_cells(r.run, r.alias, r.baseline, r.result,
+                            ':+1:' if r.accepted else ':-1:')
+        ctx.output.write(str(table))
```
tchaikov, line 116: a sample output looks like https://gist.github.com/tchaikov/8b24bdab286d65bf7cfaf0b145c031f4

compare.py
```diff
@@ -50,20 +89,38 @@ def main():
                 break
             results.extend(current.evaluate(baseline))
 
-    accepted = sum(result.accepted for result in results)
+    nr_accepted = sum(result.accepted for result in results)
```
rzarzynski, line 92: :+1:

compare.py
```diff
@@ -50,20 +89,38 @@ def main():
...
+            heading = Heading3(f'{nr_rejected} out of {nr_tests} failed')
+        else:
+            heading = Heading3(f'all {nr_tests} tests passed')
+        ctx.output.write(str(heading))
+
+        table = Table()
+        table.add_headers('run', 'metric', 'baseline', 'result', 'accepted')
+        for r in results:
+            table.add_cells(r.run, r.alias, r.baseline, r.result,
+                            ':+1:' if r.accepted else ':-1:')
```
rzarzynski, line 115: OK, so thumbs up  and thumbs down. The problem might be readability as these emoticons are pretty similar (:+1: :-1:). How about using only one of them are leaving a gap for the other?

tools/githubcheck/githubcheck/check.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class Check:
+    def __init__(self, owner, project, context):
+        self.owner = owner
+        self.project = project
+        self.context = context
+        self.github = github3.GitHub()
+
+    def login_as_app(self, pem, app_id, install_id):
```
rzarzynski, line 34: This looks actually like an obligatory, second phase of initialization. Is there any scenario where `login_as_app()` may be left uncalled? If no, then maybe merging with `__init__` could simplify the interface.

tools/githubcheck/githubcheck/check.py
```diff
@@ -0,0 +1,105 @@
...
+        completed_at = datetime.datetime.now(utc).isoformat()
+        try:
+            check_run = next(c for c in repo.commit(sha).check_runs()
+                             if c.name == self.context)
+        except github3.exceptions.GitHubException as e:
+            log.error(f"could not retrieve existing check runs for #{sha}:"
+                      f" {e}")
+        except StopIteration:
+            # create a new one
+            log.debug(f"could not find existing check runs {self.context} for "
```
rzarzynski, line 75: When this can happen? Is it allowed to `complete()` a run that has not been `start()`ed yet?

compare.py
tchaikov: how about just using :x: or :red_circle: on failure ?

rzarzynski: Makes sense!

tools/githubcheck/githubcheck/check.py
tchaikov: there are two cases, actually. 

- APIs does not needs to `login_as_app()`
- APIs requires so.

but currently, all methods exposed by `Check` uses the check API, which requires the GitHub application installation authentication. so, i am moving it into `__init__()`, will extract it out when necessary.

tools/githubcheck/githubcheck/check.py
tchaikov: i think it's just defensive coding in case that the `start()` call is not called by jenkins or just fails for any reason. but `Check` should not fail in that case.

tchaikov:
@rzarzynski updated and repushed.

tchaikov:
squashed and repushed.

PR #224, opened 2020-07-15
tools/githubcheck: pass details_url and external_id to check

tchaikov:
will need it for adding a link to jenkins job


tchaikov:
for the change on jenkins side, see https://github.com/ceph/ceph-build/pull/1618

PR #229, opened 2020-11-09
cluster/ceph: use sshtarget to find pdshhost string for mgr/mds/rgw

markhpc:
Currently sshtarget was only being used to create the pdshhost line when starting osds, but not the mgr/mds/rgw daemons.  This PR fixes that so these daemons start properly when user isn't defined.


markhpc:
Tested on o01 and properly fixes behavior when "user" isn't defined in the yaml.

PR #233, opened 2021-08-02
tools/crimson: add basic utilities to profile seastore using metrics

cyx1231st:
run.sh to run rados bench and collect results.
analyze.py to parse the results to diagram.

Signed-off-by: Yingxin Cheng <yingxin.cheng@intel.com>

### DIFF
--- a/tools/crimson/README.rst
+++ b/tools/crimson/README.rst
@@ -1,0 +1,22 @@
+=============
+Crimson tools
+=============
+
+SeaStore rados bench profiler
+=============================
+
+We can profile SeaStore using rados bench with 1 OSD.
+
+To start the test envionment:
+
+.. code-block:: console
+
+     $ MGR=0 MON=1 OSD=1 MDS=0 RGW=0 ../src/vstart.sh -n -x -d --without-dashboard --seastore --crimson --nodaemon --redirect-output
+     $ ./bin/ceph osd pool create test-pool 5 5
+     $ ./bin/ceph osd pool set test-pool size 1 --yes-i-really-mean-it
+
+Then run ``seastore_radosbench_run.sh`` to generate workload and collect
+metrics.
+
+Finally, run ``seastore_radosbench_analyze.py`` to generate diagram from the
+collected metrics.
--- a/tools/crimson/seastore_radosbench_analyze.py
+++ b/tools/crimson/seastore_radosbench_analyze.py
@@ -1,0 +1,449 @@
+#!/usr/bin/env python3
+
+from collections import defaultdict
+import json
+import os
+from os import path
+import pandas as pd
+import seaborn as sns
+
+def load_dir(dir_name):
+    load_folder = path.join(os.getcwd(), dir_name)
+    benches = []
+    metrics_start = []
+    metrics_end = []
+    for file_name in os.listdir(load_folder):
+        if not file_name.endswith(".log"):
+            continue
+        names = file_name.split("_")
+        index = int(names[1])
+        file_type = names[2]
+        file_dir = path.join(load_folder, file_name)
+
+        if file_type.startswith("bench"):
+            benches.append((index, file_dir))
+        else:
+            assert(file_type.startswith("metrics"))
+            if names[3].startswith("start"):
+                metrics_start.append((index, file_dir))
+            else:
+                assert(names[3].startswith("end"))
+                metrics_end.append((index, file_dir))
+    benches.sort()
+    metrics_start.sort()
+    metrics_end.sort()
+    return [item[1] for item in benches], [item[1] for item in metrics_start], [item[1] for item in metrics_end]
+
+def parse_bench_file(bench_file):
+    writes = 0
+    obj_size = 0
+    with open(bench_file, 'r') as reader:
+        for line in reader:
+            if line.startswith("Total writes made:"):
+                writes = int([x for x in line.split(' ') if x][3])
+            elif line.startswith("Object size:"):
+                obj_size = int([x for x in line.split(' ') if x][2])
+    assert(writes)
+    assert(obj_size)
+    return (writes * obj_size / 4096)
+
+def _load_json(file):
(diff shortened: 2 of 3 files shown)


tchaikov (approved):
probably should move it to `tools/crimson` instead of having this tool under `tools/fio_objectstore_tools`, because this tool 

1. exercises seastore using  
2. collect its metrics using the `dump_metrics` asok command.
3. process the dumped metrics and visualize them

none of these steps uses fio or talk to objectstore directly. 

also, would be great to accompany it with a tiny `README.rst` for posterities.

tools/fio_objectstore_tools/seastore/analyze.py
```diff
@@ -0,0 +1,455 @@
...
+    metrics_start.sort()
+    metrics_end.sort()
+    return [item[1] for item in benches], [item[1] for item in metrics_start], [item[1] for item in metrics_end]
+
+def parse_bench_file(bench_file):
+    writes = 0
+    obj_size = 0
+    with open(bench_file, 'r') as reader:
+        for line in reader:
+            if (line.startswith("Total writes made:")):
```
tchaikov, line 42: nit, better off dropping the parenthesis around the condition.

tools/fio_objectstore_tools/seastore/analyze.py
```diff
@@ -0,0 +1,455 @@
...
+
+    # 1. from cached_4KB, dirty_4KB
+    data_name = "cache_usage"
+
+    assert(len(dataset["cached_4KB"]) == dataset_size)
+    assert(len(dataset["dirty_4KB"]) == dataset_size)
+
+    washed_dataset[data_name] = {}
+    washed_dataset[data_name]["cached_4KB"] = dataset["cached_4KB"]
+    washed_dataset[data_name]["dirty_4KB"]  = dataset["dirty_4KB"]
```
tchaikov, line 249:
```suggestion
    washed_dataset[data_name] = {
      "cached_4KB": dataset["cached_4KB"],
      "dirty_4KB": dataset["dirty_4KB"]
    }
```

tools/fio_objectstore_tools/seastore/analyze.py
```diff
@@ -0,0 +1,455 @@
...
+
+def _load_json(file):
+    def parse_object_pairs(pairs):
+        return pairs
+    decoder = json.JSONDecoder(object_pairs_hook=parse_object_pairs)
+    with open(file, 'r') as reader:
+        line1 = next(reader).strip()
+        assert(line1.endswith("are enabled."))
+        line2 = next(reader).strip()
+        assert(line2.endswith("are enabled."))
```
tchaikov, line 58: might want to run the `ceph` cli like
```bash
$ CEPH_DEV=1 bin/ceph -s
```

to silence the warnings like:
```
2021-08-02T14:00:28.877+0800 7fb8100c9700 -1 WARNING: all dangerous and experimental features are enabled.
```

tools/fio_objectstore_tools/seastore/analyze.py
```diff
@@ -0,0 +1,455 @@
...
+    name = json_item[0]
+    value = -1
+    labels = {}
+    for k, v in json_item[1]:
+        if k == "value":
+            if isinstance(v, int):
+                value = v
+        else:
+            assert(isinstance(v, str))
+            labels[k] = v
```
tchaikov, line 74: off-topic, what about putting "labels" into its own object? so we dump a metric like
```json
{
  "name": "read-latency",
  "value": 42,
  "labels": {"shard": 0}
}
```

tools/fio_objectstore_tools/seastore/analyze.py
cyx1231st: Ah, thanks!

cyx1231st:
I'll close this one and move to `crimson/tools`

tools/fio_objectstore_tools/seastore/analyze.py
cyx1231st: How about making it single-lined, like
```
{ "name": "read-latency", "value": 42, "labels": { ... } }
```
This would be more compact to for human readers.

tchaikov: @cyx1231st i am not sure if i am following you. i am trying to propose a new scheme of how we dump labels. see https://github.com/ceph/ceph/blob/bc4aacbbbdbf52040c8e819ff0b51dee2722f06c/src/crimson/admin/osd_admin.cc#L238-L240

this is not related to how i formatted the sample json output in the comment above.

tchaikov:
> I'll close this one and move to `crimson/tools`

please just force-push to the existing remote branch to preserve the reviews and discussions.

tools/fio_objectstore_tools/seastore/analyze.py
cyx1231st: Oh, I was wondering if we can reformat the metric info into a single line. Seems we cannot if stick to the `Formatter`.

tools/fio_objectstore_tools/seastore/analyze.py
cyx1231st: all fixed

tools/fio_objectstore_tools/seastore/analyze.py
cyx1231st: all fixed

cyx1231st:
> please just force-push to the existing remote branch to preserve the reviews and discussions.

done.

tools/fio_objectstore_tools/seastore/analyze.py
tchaikov: i don't think it matters. as the json output is supposed to be consumed by a parser.  if you want to be more efficient by removing the newlines, `JSONFormatter(false)` should do the trick.

tchaikov (approved):

PR #261, opened 2021-12-31
crimson/tools: stress tool update

aisakaki:
add sequence read and write base case.

add warm up to read.

fix single core parameter doesn't work.                        
                                                               
fix node process will not be killed in the lateset ceph master.
                                                               
fix ReactorUtilizationCollectorThread will not run in the target core.


tools/crimson/crimson_stress_tool.py
```diff
@@ -703,9 +704,9 @@ def get_disk_name(self):
                     classic-memstore or classic-bluestore')
     parser.add_argument('--single-core',
             type = bool,
-            default = True,
-            help = 'run osds in single core')
-
+            default = False,
+            help = 'do not run osds in single core')
```
cyx1231st, line 708: I think it is better to use `action = 'store_true'` for the boolean options, see https://docs.python.org/3/library/argparse.html#action

So that users can use `--single-core` instead of `--single-core Ture`.

tools/crimson/crimson_stress_tool.py
```diff
@@ -91,19 +91,41 @@ def analyse(self):
...
         super().__init__(env)
-        self.iops_key = "read_iops"
-        self.latency_key = "read_latency"
-        self.bandwidth_key = "read_bandwidth"
+        self.iops_key = "sw_iops"
+        self.latency_key = "sw_latency"
+        self.bandwidth_key = "sw_bandwidth"
+    def create_command(self):
+        rados_bench_write = "sudo taskset -c " + self.task_set \
+            + " bin/rados bench -p " + self.pool + " " \
```
cyx1231st, line 109: I think it is better to be explicit about the block size for sequential workload instead of relying on the default value.

tools/crimson/crimson_stress_tool.py
```diff
@@ -125,12 +147,44 @@ def pre_process(env):
...
+        rados_bench_seq_read = "sudo taskset -c " + self.task_set \
+            + " bin/rados bench -p " + self.pool + " " \
+            + self.time + " seq -t " \
+            + str(self.thread_num) \
+            + " --no-cleanup"
+        return rados_bench_seq_read
+    
+    @staticmethod
+    def pre_process(env):
+        env_write_command = "sudo bin/rados bench -p " + env.pool + " " \
```
cyx1231st, line 173: I think it is better to be explicit about the block size for sequential workload instead of relying on the default value.

tools/crimson/crimson_stress_tool.py
```diff
@@ -644,16 +705,24 @@ def get_disk_name(self):
             type = bool,
```
cyx1231st, line 706: For
```
parser.add_argument('--block-size',
            type = str,
            default = "4096",
            help = 'data block size')
```
IIUC this only applies to random workloads. To minimize the confusion, I suggest to change this to
```
parser.add_argument('--rand-block-size',
            type = str,
            default = "4096",
            help = 'data block size for random workloads')
```

cyx1231st (commented):
Changes to the commit title / message.
* Unnecessary tabs
* crimson/tools: add sequence read and write -> tools/crimson/crimson_stress_tool: add sequential read and write
* crimson/tools: fix bugs. -> tools/crimson/crimson_stress_tool: fix bugs

tools/crimson/crimson_stress_tool.py
aisakaki: sure, this is much more convenient!

tools/crimson/crimson_stress_tool.py
aisakaki: Now the block_size can be used to decide the sequential write block size the same as the random cases. The default value for sequential cases is 4MB.

tools/crimson/crimson_stress_tool.py
```diff
@@ -88,31 +88,69 @@ def analyse(self):
...
+
+    @staticmethod
+    def pre_process(env):
+        warm_up_time = "30"
+        env_write_command = "sudo bin/rados bench -p " + env.pool \
+            + " " + warm_up_time + " write -t 30" \
+            + " -b 4096 " \
+            + "--no-cleanup"
+        os.system(env_write_command + " >/dev/null")
+        print('ceph osd warmed up.')
```
cyx1231st, line 100: Probably remove this warmup?

tools/crimson/crimson_stress_tool.py
```diff
@@ -88,31 +88,69 @@ def analyse(self):
...
         super().__init__(env)
-        self.iops_key = "read_iops"
-        self.latency_key = "read_latency"
-        self.bandwidth_key = "read_bandwidth"
-    
+        self.iops_key = "sw_iops"
+        self.latency_key = "sw_latency"
+        self.bandwidth_key = "sw_bandwidth"
+        self.block_size = '4M'
+        if env.args.block_size != "4096":
```
cyx1231st, line 118: Maybe set block_size = 0 as the special default, instead of the meaningful 4096?

tools/crimson/crimson_stress_tool.py
```diff
@@ -641,19 +731,27 @@ def get_disk_name(self):
             help = 'choose from crimson-seastore, crimson-cyanstore,\
                     classic-memstore or classic-bluestore')
     parser.add_argument('--single-core',
-            type = bool,
-            default = True,
-            help = 'run osds in single core')
-    
+            action = 'store_true',
+            default = False,
+            help = 'do not run osds in single core')
```
cyx1231st, line 736: Seems strange to use --single-core as "do not run osds in single core"

tools/crimson/crimson_stress_tool.py
```diff
@@ -199,7 +274,14 @@ def pre_process(env):
             command += " 2>/dev/null"
             os.system(command)
             env.images.append(image_name)
-    
+        print('all fio rbd images created.')
+        env_write_command = "sudo bin/rados bench -p " + env.pool \
+            + " " + warm_up_time + " write -t 30" \
+            + " -b 4096 " \
```
cyx1231st, line 280: why this is hard-coded?

tools/crimson/crimson_stress_tool.py
```diff
@@ -618,7 +707,8 @@ def get_disk_name(self):
     parser.add_argument('--block-size',
             type = str,
             default = "4096",
```
cyx1231st, line 709: Maybe it's better to set block_size = 0 as the special default, instead of the meaningful 4096?

tools/crimson/crimson_stress_tool.py
```diff
@@ -632,7 +722,7 @@ def get_disk_name(self):
             default = "result.txt",
             help = 'path of all output result after integrating')
     parser.add_argument('--output-horizontal',
-            type = bool,
+            action = 'store_true',
```
cyx1231st, line 725: for store_true, I think the default value can be removed.

tools/crimson/crimson_stress_tool.py
aisakaki: changed in new commit

cyx1231st (approved):

PR #268, opened 2022-03-22
tools/crimson/crimson_stress_tool: add fio sequence read/write case 

aisakaki:
add fio sequence read/write cases.

modified some formatting and output.

code clean up.

### DIFF
--- a/tools/crimson/crimson_stress_tool.py
+++ b/tools/crimson/crimson_stress_tool.py
@@ -105,7 +105,7 @@ class RadosSeqWriteThread(RadosRandWriteThread):
         self.iops_key = "sw_IOPS"
         self.latency_key = "sw_Latency"
         self.bandwidth_key = "sw_Bandwidth"
-        self.block_size = '4M'
+        self.block_size = env.default_seq_block_size
         if env.args.block_size != "4096":
             self.block_size = env.args.block_size
 
@@ -187,7 +187,7 @@ class RadosSeqReadThread(RadosRandWriteThread):
 
     @staticmethod
     def pre_process(env):
-        block_size = '4M'
+        block_size = env.default_seq_block_size
         if env.args.block_size != "4096":
             block_size = env.args.block_size
         env_write_command = "sudo bin/rados bench -p " + env.pool + " " \
@@ -251,8 +251,7 @@ class FioRBDRandWriteThread(Task):
                     match = re.search(r'avg=.*?,', line)
                     if match:
                         match_res = match.group()
-                        result_dic[self.lat] = float(
-                            match_res[4:-1])/1000000  # s
+                        result_dic[self.lat] = float(match_res[4:-1])/1000000  # s
                 if temp_lis[0] == "bw":
                     match_res = re.search(r'avg=.*?,', line).group()
                     result_dic[self.bw] = float(match_res[4:-1])/1000  # MB/s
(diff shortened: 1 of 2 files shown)


tools/crimson/crimson_stress_tool.py
```diff
@@ -418,8 +421,10 @@ def analyse(self):
             if temp_lis and temp_lis[0] == self.dev:
                 result_dic['Device_IPS'] = float(temp_lis[1])
                 result_dic['Device_OPS'] = float(temp_lis[7]) 
-                result_dic['Device_Read(MB/s)'] = round(float(temp_lis[2])/1024, 3)  # MB per second
-                result_dic['Device_Write(MB/s)'] = round(float(temp_lis[8])/1024, 3)  # MB per second
+                result_dic['Device_Read(MB/s)'] \
+                        = round(float(temp_lis[2])/1024, 3)  # MB per second
+                result_dic['Device_Write(MB/s)'] \
+                        = round(float(temp_lis[8])/1024, 3)  # MB per second
```
ljx023: /1024 to /1000

tools/crimson/crimson_stress_tool.py
```diff
@@ -315,6 +301,63 @@ def post_process(env, test_case_result):
...
+
+
+class FioRBDSeqWriteThread(FioRBDRandWriteThread):
+    def __init__(self, env):
+        super().__init__(env)
+        self.rw = "write"
+        self.lat = 'fio_sw_Latency'
+        self.bw = 'fio_sw_Bandwidth'
+        self.iops = 'fio_sw_IOPS'
+        if self.bs != "4096":
```
cyx1231st, line 321: Seems the logic is not correct here.

tools/crimson/crimson_stress_tool.py
```diff
@@ -315,6 +301,63 @@ def post_process(env, test_case_result):
...
+            self.bs = env.default_seq_block_size
```
cyx1231st, line 322: I suggest to expose 3 different block size configurations to user: random workload block size, sequential workload block size, and warm up block size.

tools/crimson/crimson_stress_tool.py
```diff
@@ -120,13 +120,8 @@ def create_command(self):
...
     def pre_process(env):
-        warm_up_time = "30"
-        env_write_command = "sudo bin/rados bench -p " + env.pool \
-            + " " + warm_up_time + " write -t 64" \
-            + " -b " + env.args.block_size \
-            + " --no-cleanup"
-        os.system(env_write_command + " >/dev/null")
-        print('ceph osd warmed up.')
+        # warm up for 30s
+        env.rados_pre_write(env.args.block_size, 64, 30)
```
cyx1231st, line 124: I suggest to expose 2 different time configurations to user: benchmark time and warmup time.

tools/crimson/crimson_stress_tool.py
ljx023: if fio and rados bench will not exist at same time, the "fio_" can be removed

tools/crimson/crimson_stress_tool.py
ljx023: Well, I think test block size and warm up block size are enough. Random and sequential can be distinguished and changed in their own class.

tools/crimson/crimson_stress_tool.py
```diff
@@ -251,8 +251,7 @@ def analyse(self):
                     match = re.search(r'avg=.*?,', line)
                     if match:
                         match_res = match.group()
-                        result_dic[self.lat] = float(
-                            match_res[4:-1])/1000000  # s
+                        result_dic[self.lat] = float(match_res[4:-1])/1000000  # s
```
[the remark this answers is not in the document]
aisakaki: done

tools/crimson/crimson_stress_tool.py
aisakaki: ok

tools/crimson/crimson_stress_tool.py
aisakaki: ok

tools/crimson/crimson_stress_tool.py
aisakaki: ok, now it will not support running different types of workloads in the same time such as fio rbd and rados bench.

aisakaki:
ok

cyx1231st (approved):

PR #269, opened 2022-05-11
tools/crimson/seastore_metrics_analyze: add graph for segment cleaner metrics

cyx1231st:
Parse metrics from https://github.com/ceph/ceph/pull/46158


cyx1231st:
The dependent PR has been merged.

PR #273, opened 2022-07-12
tools/crimson/seastore_metrics_analyze: rename segment_cleaner to async_cleaner

cyx1231st:
Due to https://github.com/ceph/ceph/pull/46885


cyx1231st:
@aisakaki

PR #277, opened 2023-02-12
cbt: fix prefill skips

NitzanMordhai:
after commit 43b553579f6c36b63027b31c7dbb6d0de7e46bed merge that handle prefill we are checking if folder exist after initialize and it will cause the run to skip the test.

Signed-off-by: Nitzan Mordechai <nmordec@redhat.com>

### DIFF
--- a/cbt.py
+++ b/cbt.py
@@ -74,7 +74,7 @@ def main(argv):
         for iteration in range(settings.cluster.get("iterations", 0)):
             benchmarks = benchmarkfactory.get_all(archive_dir, cluster, iteration)
             for b in benchmarks:
-                if b.exists():
+                if b.exists() and not settings.cluster.get('is_teuthology', False):
                     continue
 
                 if rebuild_every_test:


neha-ojha (approved):
Looks good to me, @markhpc can you please verify the change at your end as well

markhpc:
@NitzanMordhai Can you tell me a bit more about the condition you are hitting?  The goal with that code is to not overwrite existing output in the results directory.  Are we hitting a case where the result directory is empty but we are still skipping tests?

NitzanMordhai:
> Can you tell me a bit more about the condition you are hitting? The goal with that code is to not overwrite existing output in the results directory. Are we hitting a case where the result directory is empty but we are still skipping tests?

@markhpc After benchmark initialization (https://github.com/ceph/cbt/blob/f4acd2a39d78becde2e5663d2dde3298e0c17a3d/cbt.py#L64) we have result directory which will cause the second loop of benchmark to skip the b.run() (https://github.com/ceph/cbt/blob/master/cbt.py#L77-#L78)

Matan-B:
Just to add my two cents to the discussion, Crimson's perf suite looks good with this PR applied.
https://pulpito.ceph.com/matan-2023-03-26_14:55:21-crimson-rados:perf-main-distro-crimson-smithi/

NitzanMordhai:
@markhpc we are waiting for your approval as well.

PR #278, opened 2023-03-05
cluster/ceph: Use beast by default for radosgw

markhpc:


cluster/ceph.py
```diff
@@ -505,10 +505,10 @@ def start_rgw(self):
                 # set the rgw_frontends
                 rgw_frontends = None
                 if ssl_certificate is not None:
-                    rgw_frontends = "civetweb ssl_certificate=%s" % ssl_certificate
+                    rgw_frontends = "beast ssl_certificate=%s" % ssl_certificate
                 if port is not None:
                     if rgw_frontends is None:
-                        rgw_frontends = "civetweb"
+                        rgw_frontends = "beast"
```
cbodley, line 511: > rgw_frontends += " port=%s" % port

note that beast expects this to be option be `ssl_port={port}` for https. civetweb used to expect `port={port}s`, so this probably didn't work quite right for https in the first place. both frontends are still documented for pacific at https://docs.ceph.com/en/pacific/radosgw/frontends/

cbodley (approved):

cluster/ceph.py
markhpc: Thanks Casey!  I changed it based on your comment.

PR #279, opened 2023-03-05
cluster/ceph: split stderr/stdout logging

markhpc:


markhpc:
This worked in testing

PR #283, opened 2023-06-01
cbt: fix prefill skips

NitzanMordhai:
after commit 43b553579f6c36b63027b31c7dbb6d0de7e46bed merge that handle prefill we are checking if folder exist after initialize and it will cause the run to skip the test.
i Added also check to do the test in case it is teuthology run, we will skip all the test since we are using exist cluster.

### DIFF
--- a/cbt.py
+++ b/cbt.py
@@ -74,7 +74,7 @@ def main(argv):
         for iteration in range(settings.cluster.get("iterations", 0)):
             benchmarks = benchmarkfactory.get_all(archive_dir, cluster, iteration)
             for b in benchmarks:
-                if b.exists():
+                if b.exists() and not settings.cluster.get('is_teuthology', False):
                     continue
 
                 if rebuild_every_test:


NitzanMordhai:
@markhpc  @neha-ojha i changed the PR so we will check if its teuthology run and not skip the bench test in that case.
For some reason, i mistakenly closed the original PR https://github.com/ceph/cbt/pull/277

NitzanMordhai:
@markhpc On that run, we don't have the PR yet:
http://pulpito.front.sepia.ceph.com/nmordech-2023-06-13_04:41:43-rados:perf-main-distro-default-smithi/
None of the cbt task were running - you can see in the logs:
INFO     - cbt      - Skipping existing test in /home/ubuntu/cephtest/archive/cbt/results/00000000/id-7da5bbcd
the result's directory doesn't have any results, the benchmark was skipped..

With the PR (using force-branch):
http://pulpito.front.sepia.ceph.com/nmordech-2023-06-13_04:56:56-rados:perf-main-distro-default-smithi/
the results are in (since the benchmark did run)

cc @neha-ojha

neha-ojha (approved):

markhpc (approved):
LGTM

PR #285, opened 2023-07-07
tools/crimson: fix bunch of bugs mostly caused by update of upstream Ceph

Svelar:
tools/ crimson:
1. Add fault tolerance for stats data, iostat 11.7.3 issues fixed.
2. Fix metrics file can't be parsed, fix FIO or RADOS test can't be used.
3. Fix issues caused by metrics keys.

### DIFF
--- a/tools/crimson/seastore_metrics_analyze.py
+++ b/tools/crimson/seastore_metrics_analyze.py
@@ -120,7 +120,10 @@ def load_dir(dir_name, headcut, tailcut):
     benches = benches[headcut:tailcut]
     metrics = metrics[headcut:diff_tailcut]
     stats = stats[headcut:diff_tailcut]
-    bench_start = ret_times[0]
+    if len(ret_times) == 0:
+        bench_start = []
+    else:
+        bench_start = ret_times[0]
     bench_skip = 0
     for t in ret_times[0:headcut]:
         bench_skip += t
@@ -160,7 +163,16 @@ def _load_json(file):
     def parse_object_pairs(pairs):
         return pairs
     with open(file, 'r') as reader:
-        return json.load(reader, object_pairs_hook=parse_object_pairs)
+        names = file.split('_')
+        if names[2] == "stats":
+            return json.load(reader, object_pairs_hook=parse_object_pairs)
+        else:
+            assert(names[2] == "metrics")
+            json_dict = json.load(reader, object_pairs_hook=parse_object_pairs)
+            json_tmp = []
+            for i in json_dict[0][1]:
+                json_tmp.append(i[0])
+            return json_tmp
 
 def _process_json_item(json_item):
     name = json_item[0]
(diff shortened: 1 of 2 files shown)


cyx1231st:
* There is a wrong merge commit https://github.com/ceph/cbt/pull/285/commits/b1694296be01d666d6f3e858a1457ff06cc6fb19, which should be removed.
* Please add prefix "crimson/tools:" to the commit title, like the other merged commits.
* Looks signed-off-by is missing, please refer to https://github.com/ceph/ceph/blob/main/SubmittingPatches.rst

And thanks! I haven't updated these scripts for a while, they must be out-of-date, might need to recap.

Svelar:
> * There is a wrong merge commit [b169429](https://github.com/ceph/cbt/commit/b1694296be01d666d6f3e858a1457ff06cc6fb19), which should be removed.
> * Please add prefix "crimson/tools:" to the commit title, like the other merged commits.
> * Looks signed-off-by is missing, please refer to https://github.com/ceph/ceph/blob/main/SubmittingPatches.rst
> 
> And thanks! I haven't updated these scripts for a while, they must be out-of-date, might need to recap.

Done, review again plz, thanks.

tools/crimson/seastore_metrics_run.sh
```diff
@@ -52,7 +54,7 @@ collect_stats() {
     "value": ${read_wrtn_dscd_kb[1]}
   },
   "dscd_kb": {
-    "value": ${read_wrtn_dscd_kb[2]}
```
cyx1231st, line 55: > Add fault tolerance for stats data, iostat 11.7.3 issues fixed.

I'm not following this change. The goal was to analysis how the raw nvme SSD behaves used by SeaStore/BlueStore, which should include read, write and **discard** operations.

My local version is
```
$ iostat -V
sysstat version 12.2.0
```

My local output is:
```
iostat -k -d /dev/nvme1n1
...
Device             tps    kB_read/s    kB_wrtn/s    kB_dscd/s    kB_read    kB_wrtn    kB_dscd
nvme1n1           xx           xx          xx            xx           aa        bb          cc
```

So `iostat -k -d /dev/nvme1n1 | awk 'NR == 4 {print $6, $7, $8}'` is:
```
aa bb cc
```

which are correct.

tools/crimson/seastore_metrics_analyze.py
```diff
@@ -160,7 +163,15 @@ def _load_json(file):
     def parse_object_pairs(pairs):
         return pairs
     with open(file, 'r') as reader:
-        return json.load(reader, object_pairs_hook=parse_object_pairs)
+        names = file.split('_')
+        if names[2]=="stats":
```
cyx1231st, line 167: nit, `names[2] == "stats"`

tools/crimson/seastore_metrics_run.sh
Svelar: Perhaps it is caused by diff models of device. Added judgement.

tools/crimson/seastore_metrics_analyze.py
```diff
@@ -160,7 +163,15 @@ def _load_json(file):
...
+            return json.load(reader, object_pairs_hook=parse_object_pairs)
+        else:
```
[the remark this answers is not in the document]
Svelar: Yea

tools/crimson/seastore_metrics_analyze.py
```diff
@@ -120,7 +120,10 @@ def load_dir(dir_name, headcut, tailcut):
     benches = benches[headcut:tailcut]
     metrics = metrics[headcut:diff_tailcut]
     stats = stats[headcut:diff_tailcut]
-    bench_start = ret_times[0]
+    if len(ret_times)==0:
```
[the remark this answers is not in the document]
Svelar: done

tools/crimson/seastore_metrics_analyze.py
```diff
@@ -120,7 +120,10 @@ def load_dir(dir_name, headcut, tailcut):
...
+        bench_start = []
```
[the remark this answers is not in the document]
Svelar: and `seastore_fio_run.sh`

tools/crimson/seastore_metrics_analyze.py
Svelar: done

tools/crimson/seastore_metrics_run.sh
```diff
@@ -62,6 +72,14 @@ collect_stats() {
   }
 }
 EOT
+    else
+      echo "Error! getting parameters, please try to execute command: iostat -k -d /dev/dev-name"
```
cyx1231st, line 76: Is the order wrong? The inner `if` is checking `nand_host_sectors` which is from the nvme command.

Or probably the if doesn't need to be nested using `exit 1`.

cyx1231st (commented):
Please squash the last commit (which is only to address the review comments) into the previous related topics.

tools/crimson/seastore_metrics_run.sh
Svelar: Sorry for my mistake, modified.

PR #287, opened 2023-07-10
crimson/tools: multicore for osds test

aisakaki:


tools/crimson/crimson_stress_tool.py
```diff
@@ -764,12 +751,20 @@ def general_pre_processing(self):
 
         command += " --nodaemon --redirect-output --nolockdep"
 
-        # config ceph
-        if self.args.single_core:
+        # config bluestore op num
+        if self.args.smp:
+            if self.args.crimson and backend == "bluestore":
+                command += " -o 'crimson_alien_op_num_threads = " + \
+                        str(self.args.smp * self.args.osd) + "'"
```
cyx1231st, line 758: Doesn't need to multiply osd.

tools/crimson/crimson_stress_tool.py
```diff
@@ -764,12 +751,20 @@ def general_pre_processing(self):
...
             if not self.args.crimson:
-                self.base_result['OSD'] = "Classic-singlecore"
-                command += " -o 'ms_async_op_threads = 1 \
-                           osd_op_num_threads_per_shard = 1 osd_op_num_shards = 1'"
+                command += " -o 'osd_op_num_shards = " + \
+                        str(self.args.smp * self.args.osd) + "'"
```
cyx1231st, line 761: Doesn't need to multiply osd.

tools/crimson/crimson_stress_tool.py
```diff
@@ -789,10 +784,13 @@ def general_pre_processing(self):
                     self.tid.append(int(t))
                 line = p_tid.readline()
 
-        # find and pin osd tids
-        if self.args.single_core:
+        # config multicore for classic
```
cyx1231st, line 787: ```# all classic osds will use cpu range 0-(smp*osd)```

tools/crimson/crimson_stress_tool.py
```diff
@@ -972,9 +983,14 @@ def rados_pre_write(self, block_size, thread_num, time):
...
-                        action='store_true',
-                        help='run osds in single core')
+    parser.add_argument('--osd',
+                        type=int,
+                        default = 1,
+                        help='how many osds')
+    parser.add_argument('--smp',
+                        type=int,
+                        default = 0,
+                        help='core per osd')
```
cyx1231st, line 993: rename --taskset to --client-taskset

tools/crimson/crimson_stress_tool.py
aisakaki: addressed

tools/crimson/crimson_stress_tool.py
aisakaki: added

tools/crimson/crimson_stress_tool.py
aisakaki: renamed

cyx1231st (approved):

PR #288, opened 2023-07-12
crimson/tools: crimson tool improvement

aisakaki:
1. support task log, will save all tasks log to target directory
2. save OSD logs for each tester
3. collect multicore reactor utilization
4. introduce Time Continuous Tasks, this will help when there is multiple time points task, for example ,to collect ru at three time point, this mechanism will automatically calculate the right time points to run the task.


aisakaki:
close this due to the new PR includes these changes: https://github.com/ceph/cbt/pull/290

PR #291, opened 2023-08-24
tools/crimson: extend smp setting range and adjust log dir

aisakaki:
1. extend smp setting range
2. adjust log output
3. add group gap

### DIFF
--- a/tools/crimson/crimson_bench_tool.py
+++ b/tools/crimson/crimson_bench_tool.py
@@ -586,24 +586,21 @@ class TesterExecutor():
     def run(self, env):
         print('running...')
         tester_count = 0
-        for client_num in env.args.client:
+        for client_index, client_num in enumerate(env.args.client):
+            env.client_num = client_num
             for thread_num in env.args.thread:
-                for smp_num in env.args.smp:
-                    env.client_num = client_num
-                    env.thread_num = thread_num
-                    env.smp_num = smp_num
-                    tester_id = str(tester_count) \
-                        + ".client{" + str(client_num) \
-                        + "}" + "_thread{" + str(thread_num) \
-                        + "}" + "_smp{" + str(smp_num) + "}"
-                    env.before_run_case(tester_id)
-                    tester = Tester(env, tester_id)
-                    temp_result = tester.run()
-                    test_case_result = env.base_result.copy()
-                    test_case_result.update(temp_result)
-                    env.after_run_case(test_case_result, tester_id)
-                    self.result_list.append(test_case_result)
-                    tester_count += 1
+                env.smp_num = env.smp_list[client_index]
+                env.thread_num = thread_num
+                tester_id = f"{tester_count}.client{{{env.client_num}}}_thread" \
+                    f"{{{env.thread_num}}}_smp{{{env.smp_num}}}"
+                env.before_run_case(tester_id)
+                tester = Tester(env, tester_id)
+                temp_result = tester.run()
+                test_case_result = env.base_result.copy()
+                test_case_result.update(temp_result)
+                env.after_run_case(test_case_result, tester_id)
+                self.result_list.append(test_case_result)
+                tester_count += 1
 
     def get_result_list(self):
         return self.result_list
@@ -651,6 +648,7 @@ class Environment():
         self.testclient_threadclass_ratio_map = {}
         self.timepoint_threadclass_num_map = {}
         self.timecontinuous_threadclass_list = []
+        self.smp_list = []
         self.base_result = dict()
         self.pid = list()
         self.tid = list()
@@ -668,6 +666,12 @@ class Environment():
         self.base_result['OPtype'] = "Mixed"
         self.backend_list = ['seastore', 'bluestore', 'memstore', 'cyanstore']
         self.store = ""
+
+        if self.args.smp == [1]:
+            for _ in self.args.client:
+                self.smp_list.append(1)
+        else:
+            self.smp_list = self.args.smp
 
     def init_thread_list(self):
         # 1. add the test case based thread classes and the ratio to the dict.
@@ -984,7 +988,10 @@ if __name__ == "__main__":
                         nargs='+',
                         type=int,
                         default=[1],
-                        help='core per osd list')
+                        help='core per osd list, default to be 1, should be one-to-one \
+                              correspondence with client list if not default. But if only input \
+                              one value, this tool will automatically extend it to match the \
+                              client list')
     parser.add_argument('--bench-taskset',
                         type=str,
                         default="1-32",
--- a/tools/crimson/crimson_bench_tool.py
+++ b/tools/crimson/crimson_bench_tool.py
@@ -26,7 +26,7 @@ class Task(threading.Thread):
         self.thread_num = env.thread_num
         self.start_time = start_time
         self.result = None
-        self.log = env.args.log
+        self.log = env.log
         self.id = id #(tester_id, thread_id)
 
     # rewrite method create_command() to define the command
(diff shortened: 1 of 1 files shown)


aisakaki:
use example: `sudo ./crimson_bench_tool.py --client 2 3 5 9 18 36 71 --thread 128 --block-size 4K --bench-taskset 140-211 --store bluestore --dev /dev/nvme0n1 --rand-write 1 --time=100 --smp 1 2 4 8 16 32 64 --log log  --crimson --freq 6`

tools/crimson/crimson_bench_tool.py
```diff
@@ -1119,5 +1127,11 @@ def rados_pre_write(self, block_size, thread_num, time):
     # execute the tester in the tester matrix
     tester_executor = TesterExecutor()
     tester_executor.run(env)
-    tester_executor.output(args.output, args.output_horizontal, filters)
+
+    output_dir = ""
+    if (args.log):
```
cyx1231st, line 1132: How about simplifying the code about logging -- Assume args.log is always meaningful, so logs are always stored?

tools/crimson/crimson_bench_tool.py
```diff
@@ -1044,6 +1055,10 @@ def rados_pre_write(self, block_size, thread_num, time):
                         default = None,
                         help='directory to store logs, no log by default. Will \
                     store all tasks results and osd log and osd stdout')
+    parser.add_argument('--gap',
+                        type=int,
+                        default = 1,
+                        help='time gap between different groups')
```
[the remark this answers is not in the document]
aisakaki: this is to wait some time before start the next tests, to reserve cooling time for hardware(e.g. cpu, disk)

tools/crimson/crimson_bench_tool.py
aisakaki: Done. I set a default log directory.

tools/crimson/crimson_bench_tool.py
```diff
@@ -669,6 +669,12 @@ def __init__(self, args):
         self.backend_list = ['seastore', 'bluestore', 'memstore', 'cyanstore']
         self.store = ""
 
+        if self.args.smp == [1]:
+            for _ in self.args.client:
+                self.smp_list.append(1)
+        else:
+            self.smp_list = self.args.smp
```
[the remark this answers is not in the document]
aisakaki: addressed

tools/crimson/crimson_bench_tool.py
```diff
@@ -669,6 +669,12 @@ def __init__(self, args):
         self.backend_list = ['seastore', 'bluestore', 'memstore', 'cyanstore']
         self.store = ""
 
+        if self.args.smp == [1]:
```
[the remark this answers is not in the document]
aisakaki: yeah this make sense, modified.

tools/crimson/crimson_bench_tool.py
```diff
@@ -586,30 +585,28 @@ def __init__(self):
...
-                        + "}" + "_smp{" + str(smp_num) + "}"
-                    env.before_run_case(tester_id)
-                    tester = Tester(env, tester_id)
-                    temp_result = tester.run()
-                    test_case_result = env.base_result.copy()
-                    test_case_result.update(temp_result)
-                    env.after_run_case(test_case_result, tester_id)
-                    self.result_list.append(test_case_result)
-                    tester_count += 1
+        for client_index in range(0, len(env.args.client)):
```
tchaikov, line 588: might want to put:
```py
for client_index, client_num in enumerate(len(env.args.client)):
    env.client_num = client_num
    for thread_index, thread_num in enumerate(env.args.thread):
        # ...
```

tools/crimson/crimson_bench_tool.py
```diff
@@ -586,30 +585,28 @@ def __init__(self):
...
+            env.client_num = env.args.client[client_index]
+            for thread_index in range(0, len(env.args.thread)):
+                env.smp_num = env.smp_list[client_index]
+                env.thread_num = env.args.thread[thread_index]
+                tester_id = str(tester_count) \
+                    + ".client{" + str(env.client_num) \
+                    + "}" + "_thread{" + str(env.thread_num) \
+                    + "}" + "_smp{" + str(env.smp_num) + "}"
```
tchaikov, line 596: please use f-string.

tools/crimson/crimson_bench_tool.py
```diff
@@ -668,6 +666,22 @@ def __init__(self, args):
...
+        if len(self.args.smp) == 1:
+            for _ in self.args.client:
+                self.smp_list.append(self.args.smp[0])
+        else:
+            self.smp_list = self.args.smp
+            if len(self.args.smp) != len(self.args.client):
+                raise Exception("smp list should match the client list")
+
+        # prepare log directory
+        date = os.popen("date +%Y%m%d.%H%M%S")
```
tchaikov, line 680: might want to put
```py
with os.popen("date +%Y%m%d.%H%M%S") as date:
    # ...
```

tools/crimson/crimson_bench_tool.py
aisakaki: ok

tools/crimson/crimson_bench_tool.py
aisakaki: all addressed~

tools/crimson/crimson_bench_tool.py
aisakaki: ok

tools/crimson/crimson_bench_tool.py
aisakaki: all addressed~

tools/crimson/crimson_bench_tool.py
```diff
@@ -586,30 +584,26 @@ def __init__(self):
...
-                    tester = Tester(env, tester_id)
-                    temp_result = tester.run()
-                    test_case_result = env.base_result.copy()
-                    test_case_result.update(temp_result)
-                    env.after_run_case(test_case_result, tester_id)
-                    self.result_list.append(test_case_result)
-                    tester_count += 1
+        for client_index, client_num in enumerate(env.args.client):
+            env.client_num = client_num
+            for _, thread_num in enumerate(env.args.thread):
```
tchaikov, line 589: if we're not interested in the index, there is no need to use `enumerate` -- the regular `in` would suffice.

aisakaki: oh thanks, modified~

tools/crimson/crimson_bench_tool.py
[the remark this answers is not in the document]
cyx1231st: I think I understand why time gap is needed. But I'm still not understand the definition of group. How does this tool group the test cases, on which basis? Please specify it before using the term of group.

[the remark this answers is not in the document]
aisakaki: Oh I see. The _group_ in this tool is defined as a group of tests threads run in one ceph instance. For example, start ceph -> start 16 rados bench clients and 1 ru threads, 1 perf threads... -> stop ceph.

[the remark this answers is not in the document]
aisakaki: Or I can use _test_ here instead of _group_ is more clear?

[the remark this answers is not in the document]
cyx1231st: OK, let's use the term "test case" instead of "group".

cyx1231st (approved):
LGTM, I'll merge if there're no further comments.

PR #292, opened 2023-10-18
tools/crimson: support perf record and flamegraph for multiple cases

aisakaki:
.data file and flamegraph will be stored in task log path


aisakaki:
@cyx1231st

cyx1231st (approved):

PR #293, opened 2023-11-01
tools/crimson: add commitID and fix latency caculation error

aisakaki:
@cyx1231st @liu-chunmei

### DIFF
--- a/tools/crimson/crimson_bench_tool.py
+++ b/tools/crimson/crimson_bench_tool.py
@@ -747,9 +747,10 @@ class Environment():
         os.system("sudo rm -rf ./dev/* ./out/*")
 
         # get ceph version
-        version = self.get_version()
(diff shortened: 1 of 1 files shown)


tools/crimson/crimson_bench_tool.py
```diff
@@ -289,6 +317,9 @@ def pre_process(env):
     def post_process(env, test_case_result):
         env.remove_images()
 
+        ratio = env.testclient_threadclass_ratio_map[FioRBDSeqWriteThread]
+        test_case_result["sw_Latency"] /= int(test_case_result['Client_num']) * ratio
```
cyx1231st, line 321: Suggested a more generic fix based on the discussion.

aisakaki: addressed

cyx1231st (approved):

PR #294, opened 2023-11-02
tools/crimson: improve settings for crimson_alien_thread_cpu_cores and crimson_alien_op_num_threads

aisakaki:
@liu-chunmei @cyx1231st

### DIFF
--- a/tools/crimson/crimson_bench_tool.py
+++ b/tools/crimson/crimson_bench_tool.py
@@ -844,7 +844,12 @@ class Environment():
         self.base_result['Core'] = self.smp_num * self.args.osd
 
         # pool
-        os.system("sudo bin/ceph osd pool create " + self.pool + " 64 64")
+        os.system(f"sudo bin/ceph osd pool create {self.pool} "
+                    f"{self.args.pg} {self.args.pg}")
+        if self.args.osd < self.args.pool_size:
+            raise Exception("pool size should <= osd number")
+        os.system(f"sudo bin/ceph osd pool set {self.pool}" \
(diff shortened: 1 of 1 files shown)


tools/crimson/crimson_bench_tool.py
```diff
@@ -844,7 +844,10 @@ def general_pre_processing(self):
         self.base_result['Core'] = self.smp_num * self.args.osd
 
         # pool
-        os.system("sudo bin/ceph osd pool create " + self.pool + " 64 64")
+        os.system(f"sudo bin/ceph osd pool create {self.pool} 64 64")
+        if self.args.osd == 1:
+            os.system(f"sudo bin/ceph osd pool set {self.pool}" \
+                        f" size 1 --yes-i-really-mean-it")
```
cyx1231st, line 850: So the pool size will be the default 3 if osd is not 1?

That will still be broken if osd number is 2.

So how about making it configurable and default to 1?

aisakaki: addressed

tools/crimson/crimson_bench_tool.py
```diff
@@ -844,7 +844,10 @@ def general_pre_processing(self):
         self.base_result['Core'] = self.smp_num * self.args.osd
 
         # pool
-        os.system("sudo bin/ceph osd pool create " + self.pool + " 64 64")
+        os.system(f"sudo bin/ceph osd pool create {self.pool} 64 64")
```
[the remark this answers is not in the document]
aisakaki: addressed

cyx1231st (approved):

PR #295, opened 2023-12-04
tools/crimson: improvment about warmup, alienstore, and osd_op_num_shards

aisakaki:
@liu-chunmei 
@cyx1231st

### DIFF
--- a/tools/crimson/crimson_bench_tool.py
+++ b/tools/crimson/crimson_bench_tool.py
@@ -647,7 +647,8 @@ class Environment():
         self.smp_list = []
         self.base_result = dict()
         self.pid = list()
-        self.tid = list()
+        self.tid = list() # without alien threads
+        self.tid_alien = list()
         self.pool = "_benchtest_"
         self.images = []
         self.thread_num = -1
@@ -790,17 +791,27 @@ class Environment():
         command += " --nodaemon --redirect-output --nolockdep"
 
         # add additional crimson bluestore ceph config
+        crimson_alien_thread_cpu_cores = ''
         if self.args.crimson and backend == "bluestore":
             op_num_threads = self.smp_num
             if self.args.crimson_alien_op_num_threads:
                 op_num_threads = \
                     self.args.crimson_alien_op_num_threads[self.test_case_id]
             command += f" -o 'crimson_alien_op_num_threads = {op_num_threads}'"
-            thread_cpu_cores = self.smp_num - 1
             if self.args.crimson_alien_thread_cpu_cores:
-                thread_cpu_cores = \
+                crimson_alien_thread_cpu_cores = \
                     self.args.crimson_alien_thread_cpu_cores[self.test_case_id]
-            command += f" -o 'crimson_alien_thread_cpu_cores = 0-{thread_cpu_cores}'"
+                alien_core_range = crimson_alien_thread_cpu_cores.split('-')
+                if 0 <= int(alien_core_range[0]) <= self.smp_num-1 or \
+                    0 <= int(alien_core_range[1]) <= self.smp_num-1:
+                    raise Exception("alien threads' cores should not collide "
+                                     "with non-alien threads' cores")
+                if int(alien_core_range[1]) < int(alien_core_range[0]):
+                    raise Exception("wrong value for crimson_alien_thread_cpu_cores")
+
+            else:
+                crimson_alien_thread_cpu_cores = f"0-{self.smp_num - 1}"
+            command += f" -o 'crimson_alien_thread_cpu_cores = {crimson_alien_thread_cpu_cores}'"
         if not self.args.crimson:
             if self.args.osd_op_num_shards:
                 command += f" -o 'osd_op_num_shards = "\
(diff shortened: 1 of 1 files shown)


tools/crimson/crimson_bench_tool.py
```diff
@@ -966,39 +962,64 @@ def create_images(self):
...
                 self.command = "sudo fio" \
                     + " -ioengine=" + "rbd" \
                     + " -pool=" + pool \
                     + " -rbdname=" + image \
                     + " -direct=1" \
                     + " -iodepth=" + str(thread_num) \
-                    + " -rw=" + rw \
+                    + " -rw=write" \
                     + " -bs=" + str(bs) \
+                    + " -size=100%" \
```
liu-chunmei, line 1014: should remove size set here, since the following set size or time.

tools/crimson/crimson_bench_tool.py
```diff
@@ -802,11 +802,14 @@ def general_pre_processing(self):
...
+            if self.args.osd_op_num_shards:
+                command += f" -o 'osd_op_num_shards = "\
+                    f"{self.args.osd_op_num_shards[self.test_case_id]}'"
             else:
-                command += " -o 'osd_op_num_shards = " + \
-                    str(self.smp_num) + "'"
+                if self.smp_num <= 8:
+                    command += " -o 'osd_op_num_shards = 8'"
+                else:
+                    command += f" -o 'osd_op_num_shards = {self.smp_num}'"
```
liu-chunmei, line 815: still has another parameter, osd_op_num_threads_per_shard,  do you check code, if osd_op_num_shards has the high priority or osd_op_num_shards_ssd has high priority?

tools/crimson/crimson_bench_tool.py
```diff
@@ -1039,22 +1060,18 @@ def rados_pre_write(self, block_size, thread_num, time):
...
-                        help='warmup time for every test case, default 10s')
-    parser.add_argument('--warmup-thread-num',
+                        default="",
+                        help='warmup data block size, default equal to block size')
+    parser.add_argument('--warmup-time',
                         type=str,
-                        default="64",
-                        help='warmup thread num for every test case, default 64')
+                        default="",
+                        help='warmup time for every test case, default equal to 5 * time')
```
cyx1231st, line 1102: This description doesn't look accurate, rados and fio seem to have different strategies here.

tools/crimson/crimson_bench_tool.py
```diff
@@ -790,17 +791,19 @@ def general_pre_processing(self):
...
         if self.args.crimson and backend == "bluestore":
             op_num_threads = self.smp_num
             if self.args.crimson_alien_op_num_threads:
                 op_num_threads = \
                     self.args.crimson_alien_op_num_threads[self.test_case_id]
             command += f" -o 'crimson_alien_op_num_threads = {op_num_threads}'"
-            thread_cpu_cores = self.smp_num - 1
             if self.args.crimson_alien_thread_cpu_cores:
-                thread_cpu_cores = \
+                crimson_alien_thread_cpu_cores = \
```
cyx1231st, line 802: Can we double check in this case that if `crimson_alien_thread_cpu_cores` is set here, it should not collide with the seastar-smp range, because they are not supposed to share the same CPU in this case?

liu-chunmei: we discussed it like if not set crimson_alien_thread_cpu_cores default use 0-(smp-1), others use the input parameters.

liu-chunmei: oh, understand your meaning, check if there is overlap with seastar smp.

tools/crimson/crimson_bench_tool.py
aisakaki: yes, this setting overrides _ssd and _hdd if non-zero.

tools/crimson/crimson_bench_tool.py
aisakaki: I added some checks below

tools/crimson/crimson_bench_tool.py
```diff
@@ -1039,22 +1059,19 @@ def rados_pre_write(self, block_size, thread_num, time):
...
-    parser.add_argument('--warmup-thread-num',
+                        default="",
+                        help='warmup data block size, default equal to block size')
+    parser.add_argument('--warmup-time',
                         type=str,
-                        default="64",
-                        help='warmup thread num for every test case, default 64')
+                        default="",
+                        help='warmup time for every test case, default equal to \
+                    5 * time in rados read case or 1 * time in fio read case')
```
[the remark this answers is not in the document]
aisakaki: ok

tools/crimson/crimson_bench_tool.py
```diff
@@ -790,17 +791,27 @@ def general_pre_processing(self):
...
             command += f" -o 'crimson_alien_op_num_threads = {op_num_threads}'"
-            thread_cpu_cores = self.smp_num - 1
             if self.args.crimson_alien_thread_cpu_cores:
-                thread_cpu_cores = \
+                crimson_alien_thread_cpu_cores = \
                     self.args.crimson_alien_thread_cpu_cores[self.test_case_id]
-            command += f" -o 'crimson_alien_thread_cpu_cores = 0-{thread_cpu_cores}'"
+                alien_core_range = crimson_alien_thread_cpu_cores.split('-')
+                if 0 < int(alien_core_range[0]) < self.smp_num-1 or \
+                    0 < int(alien_core_range[1]) < self.smp_num-1:
```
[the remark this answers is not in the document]
aisakaki: indeed

cyx1231st (approved):
LGTM, wait @liu-chunmei 's approval

tools/crimson/crimson_bench_tool.py
```diff
@@ -1150,11 +1228,27 @@ def rados_pre_write(self, block_size, thread_num, time):
                         help='set crimson_alien_op_num_threads. \
                             Equal to smp number by default.')
     parser.add_argument('--crimson-alien-thread-cpu-cores',
+                        nargs='+',
+                        type=str,
+                        default=None,
+                        help='set crimson_alien_thread_cpu_cores. The input should be ranges. \
+                            Such as 0-3 2-6. 0-smp number-1 by default.')
```
cyx1231st, line 1235: Nit, also add "For Crimson with AlienStore, alien-store-tp, bstore_aio and rocksdb threads will also be binded to this CPU range." here.

cyx1231st, line 1235: Hmm, setting absolute range here means that this is not extendable to multiple OSDs, right?

Might be worthing to add a TODO here.

aisakaki: Do you means to set range dependently for every OSD? Now all osds' alien threads will be set in the crimson_alien_thread_cpu_cores range, and also all osds will run in --smp core range

cyx1231st: OK, so when there are multiple OSDs, we can set a crimson_alien_thread_cpu_cores range to contain all of their alienstore threads.

PR #297, opened 2023-12-27
tools/crimson: introduce crimson auto bench tool

aisakaki:
Write yaml file and run ./crimson_auto_bench.py to run multiple tests according to the yaml file and analyse results and draw graphics. 
Use --run to run the whole process, --bench to only run tests, --ana to organize data and draw graphics for exist bench results.
Use --repeat to decide the repeat times for every test.
All repeat results will be drawed as a point in the result graphics, and the mean value will be drawed as a mean line.   
By default, each tests configed in the yaml file are draw in dependent graphics, if you want to compare multiple tests in the config file, use --comp, and all tests will be drawed in one graphics. But be careful, when using --comp, you must write the same --x target config in config file.
You can set --x and multiple --y to decide what index the target graphics based on.
And also, all organized data will be stored in the graphics directory.
--help for more detail.
All failed tests(exceed retry limits of crimson_bench_tool) will be recorded at failure_log.txt
You can use tree root directory to see the structure
This tool require crimson_bench_tool.py at the same directory.

### DIFF
--- a/tools/crimson/bench_config.yaml
+++ b/tools/crimson/bench_config.yaml
@@ -1,0 +1,20 @@
+# this is an example for the config file of crimson_auto_bench
+
+rand_write: 1
+client: 1 2
+block_size: 4096
+smp: 1 2
+time: 5
+tolerance_time: 5
+retry_limit: 2
+---
+fio_rbd_rand_write: 1
+client: 2 4
+block_size: 4096
+smp: 1 2
+time: 5
+tolerance_time: 5
+retry_limit: 2
+crimson: True
+store: bluestore
+iostat: 6
--- a/tools/crimson/crimson_auto_bench.py
+++ b/tools/crimson/crimson_auto_bench.py
@@ -1,0 +1,303 @@
+#!/usr/bin/env python3
+import yaml
+import argparse
+import os
+import pandas as pd
+import matplotlib.pyplot as plt
+
+''' directory structure:
+    autobench.{date}/rep:{repeat_id}/test:{test_id}
+    test:{test_id} is the output directory from crimson_bench_tool.
+
+    graphic directory:
+    graphic.autobench.{date}/test:{test_id}
+'''
+
+no_value_attributes= ['crimson', 'output_horizontal', 'perf', 'perf_record']
+
+# transfer to crimson_bench_tool param
+def trans(param):
+    res = f"--{param}"
+    res = res.replace('_', '-')
+    return res
+
+def prefix_match(prefix, target):
+    if target[:len(prefix)] == prefix:
+        return True
+    else:
+        return False
+
+def read_config(config_file, x = None, comp = None):
+    f = open(config_file, 'r')
+    _config = yaml.safe_load_all(f.read())
+    f.close()
+
+    configs = list()
+    for config in _config:
+        configs.append(config)
+
+    if x:
+        for config in configs:
+            if args.x not in config:
+                raise Exception("Error: x param should exist in config yaml file")
+            if len(config[args.x]) <= 1:
+                raise Exception("Error: cannot use single param as x axis")
+    if comp:
+        if not x:
+            raise Exception("Error: must input --x when using --comp")
+        same_x = configs[0][x]
+        for config in configs:
+            if config[x] != same_x:
+                raise Exception("Error: x must be the same "\
+                                "between different test in --comp mode")
+    return configs
+
+def do_bench(configs, repeat):
+    # all files' root for this auto bench test
+    root = ''
+    with os.popen("date +%Y%m%d.%H%M%S") as date:
+        line = date.readline()
+        res = line.split()[0]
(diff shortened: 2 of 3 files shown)


aisakaki:
@cyx1231st  @liu-chunmei

tools/crimson/bench_config.yaml
```diff
@@ -0,0 +1,20 @@
...
+fio_rbd_rand_write: 1
+client: 2 4
+block_size: 4096
+smp: 1 2
+time: 5
+tolerance_time: 5
+retry_limit: 2
+crimson: True
+store: bluestore
+iostat: 6
```
cyx1231st, line 20: Looks EOF is wrong.

tools/crimson/crimson_auto_bench.py
```diff
@@ -0,0 +1,301 @@
...
+        results = read_results(args.root)
+        res_path = f"{res_path_prefix}.{args.root}"
+        delete_and_create_at_local(res_path)
+        for y in args.y:
+            analysed_results = adjust_results(results, y)
+            draw(analysed_results, configs, args.x, y, res_path, args.comp)
+
+    if args.clean:
+        os.system("sudo rm -rf autobench.*")
+        os.system("sudo rm -rf graphic.autobench.*")
```
cyx1231st, line 301: Looks EOF is wrong.

tools/crimson/crimson_bench_tool.py
```diff
@@ -1304,9 +1315,10 @@ def check_failure(self):
                         help='how many osds')
     parser.add_argument('--log',
                         type=str,
-                        default = "log",
-                        help='directory to store logs, ./log by default. Will \
-                    store all tasks results and osd log and osd stdout')
+                        default = None,
+                        help='directory prefix to store logs, ./log_date by default.\
+                    will add _crimson/classic_backend_osd_poolsize to be log dir name. \
```
cyx1231st, line 1320: Looks it is out-of-date?

tools/crimson/crimson_auto_bench.py
```diff
@@ -0,0 +1,301 @@
...
+                        help= "remove all history graphic and bench results")
+
+    parser.add_argument('--config',
+                        type=str,
+                        default="bench_config.yaml",
+                        help="bench config yaml file path, bench_config.yaml by default")
+    parser.add_argument('--root',
+                        type=str,
+                        default=None,
+                        help="required if --ana")
```
cyx1231st, line 247: Looks --root and --ana can be merged.

tools/crimson/crimson_auto_bench.py
```diff
@@ -0,0 +1,301 @@
...
+
+    parser.add_argument('--repeat',
+                        type=int,
+                        default=1,
+                        help="repeat time for every tests, default to be 1")
+    parser.add_argument('--x',
+                        type=str,
+                        help="x axis of result graphics, required when --ana and --run")
```
cyx1231st, line 255: Is there an example of what x could be, or where to identify the possible values of x ?

tools/crimson/crimson_auto_bench.py
```diff
@@ -0,0 +1,301 @@
...
+    parser.add_argument('--y',
+                        nargs='+',
+                        type=str,
+                        default=["IOPS"],
+                        help="y asix of result graphics, IOPS by default")
```
cyx1231st, line 260: Looks y is only for the labeling purpose ?

aisakaki: yes, all results from crimson bench tool will be recorded

tools/crimson/crimson_auto_bench.py
aisakaki: x can be smp, client, thread, osd_op_num_shards, ms_async_op_threads, etc.
all the parameter that can be multiple in the crimson bench tool can be x. Beacuse x is the main variable in the target graphics to draw x-y graphics

cyx1231st: Let's elaborate them in the help message.

tools/crimson/crimson_auto_bench.py
cyx1231st: Then probably this is better:
"the label name of y asix of the result graphics, IOPS by default"

tools/crimson/crimson_bench_tool.py
aisakaki: This means the default log prefix is "log_{date}" if there is no input.

tools/crimson/crimson_auto_bench.py
aisakaki: done

tools/crimson/crimson_auto_bench.py
aisakaki: fixed

cyx1231st (approved):

PR #302, opened 2024-01-22
tools/crimson: misc improvements and fixes on task control

aisakaki:
remove all sudos
move reset_task_done() to a more appropriate position
fix pre write will not reset_task_done()
fix timepoint case will not check tester failed when waitting
fix fio(abnormal exit) not been killed in general preprocess
avoid non client task to set_task_done()


cyx1231st:
Can you split independent fixes into multiple commits ?

aisakaki:
> Can you split independent fixes into multiple commits ?

done

cyx1231st (approved):

PR #306, opened 2024-05-02
Extend librbdfio.py with a new workloads block. Add minor fixes.

perezjosibm:
This is my first PR, please be gentle in the review. We add the following new features into the CBT for the benchmark librbdfio:

-  Extend the test plan .yml with a new block section 'workloads' to allow a list of performance tests to be indicated, each with its own range of iodepth and numjobs to iterate. An example of this is provided in `example/rbd_fio_test.yml`

- Extend librbdfio.py with support for such 'workloads' block.

### New workloads block
The value of this new workloads block is to allow a finer description of a test job, for example, indicating a range for the iodepth and numjobs FIO parameters, as well as whether to monitor the execution (eg. with perf, top, or collectl). This allow the collection of measurements for computing a response curve.
Here is a snippet that describes:

- a precondition phase consisting on a random write workload, that is not being monitored, then
- a random read workload over the ranges indicated for the FIO parameters numjobs and iodepth
- a random write workload. 

 ```yaml
  workloads:
     precond1:
       jobname: 'precond1rw'
       mode: 'randwrite'
       numjobs: [ 1 ]
       iodepth: [ 8 ]
       monitor: False
     test1:
       jobname: 'rr'
       mode: 'randread'
       numjobs: [ 1, 4, 8 ]
       iodepth: [ 1, 4, 16]
     test2:
       jobname: 'rw'
       mode: 'randwrite'
       numjobs: [ 4 ]
       iodepth: [ 16 ]
```

The following snippet shows the execution of the precondition phase:
Screenshot 2024-05-07 at 16 07 38

The following snippet shows the execution of the tests1 and 2:
Screenshot 2024-05-07 at 16 08 14

The following snippet shows the resulting files from the execution of the test:
Screenshot 2024-05-13 at 11 56 20

PR #307, opened 2024-06-04
Add fio-parse-jsons.py tool to parse a set of json FIO files.

perezjosibm:
This is the separate commit only for the new standalone tool as described in https://github.com/ceph/cbt/pull/306.

### New standalone script `fio-parse-jsons.py`

This script processes a list of FIO JSON output files and produces as output:

- a gnuplot script, 
-  a .dat file with the data to plot,
-  a summary table of FIO results in wiki format, printed to stdout.

This is especially useful to produce a response graph from a set of executions ranging the number of FIO jobs and the iodepth values. The script was written before knowledge of CBT was gained, so in a way is independent of the script driving the tests. A future PR would integrate the functionality of this standalone script with that of CBT.

The following is an example of the execution of the script:

```bash
 # python3 /cbt/tools/fio-parse-jsons.py -c crimson200gb_1procs_randwrite_list -t 'Crimson 200GB RBD 4k rw' -a crimson4cores_200gb_1img_4k_1procs_randwrite_avg.json
```
the arguments are:

-  `-c config_file`: a txt file containing the list of FIO output JSON file to process,
-  `-t title`: the string ot use as title for the gnuplot chart,
- `-a cpu_avg.json`: a .json file containing the avg CPU utilisation, normally produced by the script `parse-top.pl`.

 The following are the .dat and gnuplot files produced:
```bash 
 crimson200gb_1procs_randwrite.dat
 crimson200gb_1procs_randwrite.plot
``` 

 To produce the chart, simply execute
```bash
gnuplot classic200gb_1procs_randwrite.plot
```

the IOPs vs latency chart result is shown below:

crimson200gb_1procs_randwrite

### DIFF
--- a/tools/fio-parse-jsons/README.md
+++ b/tools/fio-parse-jsons/README.md
@@ -1,10 +1,10 @@
 # fio-parse-jsons.py - a FIO post processing tool.
 
-## Description
+## Description:
 
 This is a standalone tool to assist the post processing of JSON outout files from CBT when running the FIO benchmark.
 
-The execution of the script produces as outp:
+The execution of the script produces as output:
 
 1. a gnuplot script,
 2. a .dat file with the data to plot,
@@ -13,6 +13,10 @@ The execution of the script produces as outp:
 This is especially useful to produce a response graph from a set of executions ranging the number of FIO jobs and the iodepth values.
 The script was written before knowledge of CBT was gained, so in a way is independent of the script driving the tests.
 A future PR would integrate the functionality of this standalone script with that of CBT.
+
+## Requirements:
+
+Besides the yaml and xml Python modules that CBT already depends upon, you need to install the package gnuplot according to your Linux distro.
 
 ## Usage:
 
@@ -29,6 +33,7 @@ the arguments are:
 - `-a cpu_avg.json:` a .json file containing the avg CPU utilisation, normally produced by the script parse-top.pl.
 
 The following are the .dat and gnuplot files produced:
+
 ```bash
  crimson200gb_1procs_randwrite.dat
  crimson200gb_1procs_randwrite.plot
@@ -41,3 +46,6 @@ gnuplot classic200gb_1procs_randwrite.plot
 ```
 
 the IOPs vs latency chart result is shown below:
+
+![crimson200gb_1procs_randwrite](https://github.com/ceph/cbt/assets/23522684/44aeeb17-b99f-48c7-bd0e-3a443c2e5d90)
+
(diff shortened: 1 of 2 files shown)


tools/fio-parse-jsons/README.md
```diff
@@ -0,0 +1,43 @@
...
+ crimson200gb_1procs_randwrite.plot
+```
+
+To produce the chart, simply execute
+
+```bash
+gnuplot classic200gb_1procs_randwrite.plot
+```
+
+the IOPs vs latency chart result is shown below:
```
sseshasa, line 43: The link to the sample charts need to be added here.

tools/fio-parse-jsons/README.md
```diff
@@ -0,0 +1,43 @@
+# fio-parse-jsons.py - a FIO post processing tool.
+
+## Description
+
+This is a standalone tool to assist the post processing of JSON outout files from CBT when running the FIO benchmark.
+
+The execution of the script produces as outp:
```
sseshasa, line 7: typo: s/outp/output

tools/fio-parse-jsons/README.md
```diff
@@ -0,0 +1,43 @@
+# fio-parse-jsons.py - a FIO post processing tool.
```
sseshasa, line 1: The documentation must also include information about dependencies  to install for the tool. For e.g., gnuplot, python modules (if any) etc..

sseshasa (approved):
Just added a couple of comments related to the documentation. I just took a cursory look at the tool and in general it looks good to me.

tools/fio-parse-jsons/README.md
perezjosibm: Makes sense, thanks I'll add these details

tools/fio-parse-jsons/README.md
perezjosibm: Yes, sorry I missed that 😃

perezjosibm:
Hi @sseshasa, I've updated the PR as per your comments, would you mind to merge this PR (since this being my first one I don't have write permissions yet) Thanks!

sseshasa:
> Hi @sseshasa, I've updated the PR as per your comments, would you mind to merge this PR (since this being my first one I don't have write permissions yet) Thanks!

@perezjosibm Sure! Just FYI, when a change is requested on a commit, it is better to modify the same commit instead of adding a new one. You can make the requested changes and either rebase it on the existing commit itself or use "git commit --amend" to update the last commit like in this case. This helps in maintaining a good clean git history.

The above is something to keep in mind for next time. I will go ahead merge this.

perezjosibm:
Many thanks @sseshasa, yes I was wondering about that, I'll amend each time I got a new commit then 👍 Cheers

PR #314, opened 2024-08-02
Create a new rbd_fio_test.yaml that demos how to create a response curve

lee-j-sanders:
Create a new rbd_fio_test.yaml that demos how to create a response curve using the PR306 tooling that was added to CBT.

I had to delete the existing YML because it contained a small amount of sensitive data and create a new YAML.
I also used the opportunity to change the name to follow the same file extension as the other YAMLs (.yaml instead of .yml).

I've tested these changes with the example response curve that I'll attach to the PR.

Users of this PR, will need to change the mysystem.mycompany.com to something more appropriate to their system.


lee-j-sanders:
Example graph here:

4cores_8img_1procs_randwrite_iops_vs_lat

lee-j-sanders:
Mark please review my PR. Thanks!

PR #318, opened 2024-10-11
cbt: creating a common base class for fio subclasses

harriscr:
This is the first part of an effort to rationalise the 5 current fio classes into a single class structure. It will mean the users of cbt no longer need to know the subtle differences between the differing fio classes.
The change adds typing for methods and variables (using mypy) and pep8 compliance (using ruff) to the fio* files changed.

This is my first pull request so I expect there is something I've missed in the process.

This has been tested by:
1. running the unit tests in the test sub directory
2. Validating that the client_endpoint_rbd_4K_rand_write.yaml configuration file used in teuthology results in the same generated fio cli with and without these changes
3. Validating that the example yaml for fio (example-client_endpoints.yaml) generated the same cli to run fio with and without the changes


benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
+    def fio_command_extra(self, endpoint_number: int) -> str:
+        command: str = ""
 
         # typical directory endpoints
-        if self.endpoint_type == 'directory':
-            for proc_num in range(self.procs_per_endpoint):
-                cmd += ' --name=%s/`%s`-%s-%s' % (self.endpoints[ep_num], common.get_fqdn_cmd(), ep_num, proc_num)
+        if self._endpoint_type == "directory":
+            for proc_num in range(self._procs_per_endpoint):
+                command = (
```
perezjosibm, line 78: I think the semantics of the new command is different to the original construction cmd: the original constructs a series of strings with option "--name+<path>" so we have one per proc_num, but the new command only retains the latter, so all the previous entries are lost.

benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
-        cmd += ' --iodepth=%d' % self.prefill_iodepth
-        cmd += ' --size %dM' % self.size
-        cmd += ' --output-format=%s' % self.fio_out_format
-        cmd += self.fio_command_extra(ep_num)
-        return cmd
-
-    def prefill(self):
-        super(Fio, self).prefill()
-        if not self.prefill_flag:
+        elif self._endpoint_type == "rbd":
```
perezjosibm, line 84: This is interesting: suppose that your actual workload test run is using librbdfio, but you would like to run the prefill directly via libaio, which currently in CBT is using FIO raw. However, my impression so far is that the endpoints are configured to use the same engine as the one for running the actual workload. 
Shall we consider to have the flexibility of using different FIO engines for the prefill?
@markhpc you thoughts please

benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
+            pool_name, rbd_name = self._endpoints[endpoint_number].split("/")
+            command += " --clientname=admin --invalidate=0"
+            command += f" --pool={pool_name} --rbdname={rbd_name}"
+            for proc_num in range(self._procs_per_endpoint):
+                rbd_name = f"{self._endpoints[endpoint_number]}-{proc_num}"
+                command += f" --name={rbd_name}"
+        return command
+
+    def _build_prefill_command(self, endpoint_number: int) -> str:
+        command = f"sudo {self.cmd_path} --rw=write --bs=4M --iodepth={self._configuration_with_defaults.get('prefill_iodepth')}"
```
perezjosibm, line 94: Here is assuming that prefill can only be done using block size of 4MB, which I don't object per se, but wondering whether that could be the default value but to allow the flexibility of different block sizes?

benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
+            if self._recovery_test_type == "blocking":
                 recovery_callback = self.recovery_callback_blocking
-            elif self.recov_test_type == 'background':
+            elif self._recovery_test_type == "background":
                 recovery_callback = self.recovery_callback_background
-            self.cluster.create_recovery_test(self.run_dir, recovery_callback, self.recov_test_type)
+            self._cluster.create_recovery_test(self.run_dir, recovery_callback, self._recovery_test_type)  # type: ignore [no-untyped-call]
 
-        if 'recovery_test' in self.cluster.config and self.recov_test_type == 'background':
+        if "recovery_test" in self._cluster.config and self._recovery_test_type == "background":
```
perezjosibm, line 136: Probably this block of code requires its own method

benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
+        monitoring.start(self.run_dir)  # type: ignore [no-untyped-call]
 
-        logger.info('Running fio %s test.', self.mode)
-        ps = []
-        for i in range(self.endpoints_per_client):
-            p = common.pdsh(settings.getnodes('clients'), self.run_command(i))
-            ps.append(p)
-        for p in ps:
-            p.wait()
+        log.info("Running fio %s test.", self._cli_options["rw"])
```
perezjosibm, line 142: I think the original 'self.mode' was more generic to refer to the IO mode (random read, random write, etc), whereas the suggested '._cli_options["rw"]' is too specific to FIO, so not sure if that conveys the intention.
If in the near future we need to compare two different benchmarks over the same workload, we need that the options generic enough to compare apples with apples, so my vote would be towards a generic IO mode rather than specific FIO "rw".
@markhpc , @sseshasa what are your thoughts please?

benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
 
-        monitoring.stop(self.run_dir)
+        monitoring.stop(self.run_dir)  # type: ignore [no-untyped-call]
 
         # Finally, get the historic ops
-        self.cluster.dump_historic_ops(self.run_dir)
-        common.sync_files('%s/*' % self.run_dir, self.out_dir)
-        self.analyze(self.out_dir)
-
-    def cleanup(self):
```
perezjosibm, line 251: |I think its a good opportunity to add method documentation

benchmark/fio.py
```diff
@@ -1,273 +1,178 @@
...
+    def recovery_callback_background(self) -> None:
+        log.info("Recovery thread completed!")
 
-    def analyze(self, out_dir):
-        logger.info('Convert results to json format.')
-        for client in settings.getnodes('clients').split(','):
-            host = settings.host_info(client)["host"]
-            for i in range(self.endpoints_per_client):
+    def analyze(self, output_directory: str) -> None:
+        log.info("Converting results to json format.")
```
perezjosibm, line 167: Method documentation?

perezjosibm:
Adding @sseshasa as well since the refactoring changes seem important

perezjosibm:
@harriscr can you kindly post the snapshots showing the tests that you run? Also, it would be very useful the class diagrams that you got from `pyreverse` (new vs old).

I have some problems to understand what you mean by:

> It will mean the users of cbt no longer need to know the subtle differences between the differing fio classes.

I'm guessing the main difference between each FIO subclass instance would be the underlying FIO engine (eg.libaio, librd, etc.) Or you mean "block" vs 'file" IO? Could you illustrate with an example please?

Thanks

perezjosibm (changes requested):
I've added comments to the points I consider might need amending/clarification. It looks good to me so far, but will be useful to have further insight from Mark and Seshasayee.
Thanks

benchmark/benchmark.py
```diff
@@ -1,39 +1,39 @@
...
         self.config = config
         self.cluster = cluster
         hashable = json.dumps(sorted(self.config.items())).encode()
         digest = hashlib.sha1(hashable).hexdigest()[:8]
-        self.archive_dir = os.path.join(archive_dir,
-                                        'results',
-                                        '{:0>8}'.format(config.get('iteration')),
-                                        'id-{}'.format(digest))
+        self.archive_dir = os.path.join(
+            archive_dir, "results", "{:0>8}".format(config.get("iteration")), "id-{}".format(digest)
```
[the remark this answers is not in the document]
harriscr: I agree. I plan to look at benchmark.py in the future, so will look at changing this line to an f-string when I do.

benchmark/benchmark.py
```diff
@@ -1,39 +1,39 @@
...
-        self.osd_ra = config.get('osd_ra', '0')
-        self.cmd_path = ''
-        self.valgrind = config.get('valgrind', None)
-        self.cmd_path_full = ''
-        self.log_iops = config.get('log_iops', True)
-        self.log_bw = config.get('log_bw', True)
-        self.log_lat = config.get('log_lat', True)
+        logger.info("Results dir: %s", self.archive_dir)
+        self.run_dir = os.path.join(
+            settings.cluster.get("tmp_dir"), "{:0>8}".format(config.get("iteration")), self.getclass()
```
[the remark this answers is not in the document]
harriscr: I am working on a pyproject.toml file which will contain the formatting instructions, as well as linting and typing. I would like to deliver under a separate PR however.

benchmark/benchmark.py
```diff
@@ -130,27 +135,27 @@ def prefill(self):
...
         if not os.path.exists(self.archive_dir):
             os.makedirs(self.archive_dir)
         if not os.path.exists(config_file):
             config_dict = dict(cluster=self.config)
-            with open(config_file, 'w') as fd:
+            with open(config_file, "w") as fd:
                 yaml.dump(config_dict, fd, default_flow_style=False)
 
-    def exists(self):
+    def exists(self) -> bool:
```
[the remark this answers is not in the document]
harriscr: Correct. I will type the methods as I need to edit/change them. 
This one was changed so the signature matches in the child class and the parent class, as I was changing the child class.

benchmark/fio.py
harriscr: The idea of this particular PR is not to change the current behaviour. It is something to consider for the future however

benchmark/fio.py
harriscr: This matches the existing behaviour. self.mode is taken from the options in the yaml file, and is used for the -rw parameter to fio. The original code is:
<pre>self.mode = config.get('mode', 'write')</pre>
and the new code is:
<pre>self._cli_options["rw"] = self._configuration_with_defaults.get("mode", "write")</pre>

I'm happy to change the code to take the value from the configuration instead if you think that would be a better approach, but it doesn't change what is printed to the screen, or what the code does

benchmark/fio.py
harriscr: Can't believe I missed this. Will add in.
I'm surprised that the linter didn't pick this up either so may take a look at my settings

benchmark/fio.py
harriscr: Will add

harriscr:
teuthology runs are:
perf-basic - all passed
https://pulpito.ceph.com/harriscr-2024-11-05_18:48:16-perf-basic-wip-harriscr-cbt-refactor-2024-10-17-0839-distro-default-smithi/

rados/perf - all passed
https://pulpito.ceph.com/harriscr-2024-11-05_18:51:09-rados:perf-wip-harriscr-cbt-refactor-2024-10-17-0839-distro-default-smithi/

crimson-rados/perf - 2 failures - 'hit max job timeout' when running radosbench, which is not effected by these changes
https://pulpito.ceph.com/harriscr-2024-11-05_18:49:41-crimson-rados:perf-wip-harriscr-cbt-refactor-2024-10-17-0839-distro-default-smithi/

harriscr:
Here are the pyreverse diagrams:
classes_after
classes_before
packages_after
packages_before

PR #321, opened 2024-12-13
Add tools and code to produce hockey stick curves for a test run from a PR 319 format json results file

harriscr:
This PR is the second piece of work to get to the point where we can automatically produce plots from a CBT run.
At the moment it will only work with fio test runs, but the plan is to extend it in future to work with any set of CBT results

It follows on from [PR 319](https://github.com/ceph/cbt/pull/319) and uses the common format file as a basis for each plot.

The code is structured in such a way that it can be called from existing CBT code, or run as a stand-alone python script, which is included in this PR.

### New dependencies
The code uses the [matplotlib.pyplot](https://matplotlib.org/) library to generate the graphs, so this has been added to the requirements.txt file

The included python script can be used to plot multiple curves on a single image.
This can be 2 or more specified json files in the PR 319 format, in which case a comparison of the individual runs will be plotted.
If 2 or more directories of results are specified then a plot will be created for each common set of data in the given directories.
e.g. The 4K random read data from each directory will be plotted on a single curve.

## Example single plots
| | |
|-|-|
|4096B_randwrite | 65536B_70_30_randrw |

## Example comparison plots
|Script Invocation |Generated Plot |
|-|-|
|~/cbt/tools/plot_comparison.py --files="/tmp/main_2024_06_19/visualisation/16384B_70_30_randrw.json,/tmp/main_2024_06_19/visualisation/65536B_70_30_randrw.json" --labels="test_branch,main squid" --output_directory="/tmp/graph_examples"|Comparison_4096B_randwrite_8192B_randwrite|
|~/cbt/tools/plot_comparison.py --files="/tmp/main_2024_06_19/visualisation/16384B_70_30_randrw.json,/tmp/main_2024_06_19/visualisation/65536B_70_30_randrw.json" --labels="16K test_branch,64K main squid" --output_directory="/tmp/graph_examples"|Comparison_16384B_70_30_randrw_65536B_70_30_randrw|
|~/cbt/tools/plot_comparison.py --files="/tmp/main_2024_06_19/visualisation/32768B_read.json,/tmp/main_2024_06_19/visualisation/32768B_write.json" --labels="32K read test,32K write test" --output_directory="/tmp/graph_examples"|Comparison_32768B_read_32768B_write| |

## Example directory comparison plots
script invocation: 
[...]

PR #319, opened 2024-11-22
Create a intermediate file format from a set of fio results

harriscr:
This PR is the foundation for the work to get CBT to automatically produce a hockey-stick output curve of IOps or Bandwidth vs Latency.
It consists of a python class that will convert the json output from an fio benchmark run in CBT to a common intermediate format, the unit tests for that class, and a wrapper around the class allowing it to be run independently.

The vision is that eventually there will be a class or classes to convert the results from any CBT benchmark run into this intermediate format. This will then allow a single class/utility to be written to produce hockey-stick curves from any CBT benchmark run results.

The next steps are to write the class to plot the data, and then add the code to CBT so it gets run as part of a fio benchmark, possibly using an option in the CBT yaml file

# Testing:
Both manual and automated testing were done
### Teuthology performance tests that run CBT. This is just a regression as no code or behaviour has changed:
**perf-basic**
https://pulpito.ceph.com/harriscr-2024-11-19_09:33:42-perf-basic-wip-harriscr-cbt-refactor-2024-10-17-0839-distro-default-smithi/

**rados/perf**
https://pulpito.ceph.com/harriscr-2024-11-19_09:23:32-rados:perf-wip-harriscr-cbt-refactor-2024-10-17-0839-distro-default-smithi/
 - failed with missing test module: 
 > No module named 'tasks' in run_tasks.py

**crimson-rados/perf**
https://pulpito.ceph.com/harriscr-2024-11-19_09:34:16-crimson-rados:perf-wip-harriscr-cbt-refactor-2024-10-17-0839-distro-default-smithi/
 - failed 2 tests with job timeout: 
 > hit max job timeout
 
 ### Python unit tests:
> ======================== 326 passed, 3 skipped in 0.52s ========================
>                                                           Finished running tests!
 
 ### Python tools
 black, mypy and ruff were all run to check for pep8 compliance:
>  harriscr@eilistraee:\~/cbt $ source \~/python_environments/ceph_python/bin/activate
> (ceph_python) harriscr@eilistraee:\~/cbt $ ruff check common_output_formatter.py 
> All checks passed!
> (ceph_python) harriscr@eilistraee:\~/cbt $ mypy common_output_formatter.py
> Success: no issues found in 1 source file
> (ceph_python) harriscr@eilistraee:\~/cbt $ black common_output_formatter.py
> All done! ✨ 🍰 ✨
> 1 file left unchanged.
 
 ## Manual
 Ran the wrapper against a folder of fio test results:

[...]

PR #322, opened 2025-01-10
Add github markdown and pdf report generation for an fio test run using CBT

harriscr:
This PR is the final part of the work to generate a report for an fio test run. It generates a github markdown file and a pdf file that have the same content.

It builds on the work in [PR 319](https://github.com/ceph/cbt/pull/319) and [PR 321](https://github.com/ceph/cbt/pull/321)

The report file is initially written in markdown using the [mdutils](https://pypi.org/project/mdutils/) package, and then converted to pdf using [pandoc](https://pandoc.org/).
mdutils has been added to the requirements.txt file for cbt

Two python scripts that can be used to create the simple or comparison reports have been added to the tools directory, and these were used in testing the changes.

## Simple Report
The simple report generates a report for a single test run. It consists of a table containing the maximum throughput and latency and a set of hockey-stick curves for each individual test e.g. 4K sequential read, 4k sequential read/write etc
Finally the cbt_config.yaml file for the run is included, with host names and IP addresses redacted due to privacy concerns.

### Example simple report

[performance_report_250110_165240.pdf](https://github.com/user-attachments/files/18380087/performance_report_250110_165240.pdf)

[performance_report_250110_165240.md](https://github.com/user-attachments/files/18380141/performance_report_250110_165240.md)

The markdown report contains relative links to the contained plot files so that if the report directory is uploaded to a github repository the markdown can be viewed. 

## Comparison Report
The comparison report contains a comparison of 2 or more test run results. There is theoretically no limit to the number of results that can be compared, however the graphs are limited to 6 unique colours and in practice it has been found that after 4 the pdf table headings become somewhat garbled.

The tables for a 2-way comparison contain the maximum throughput and latency values, and the percentage difference from the baseline. For a 3 or more way report only the percentage difference in throughput is included in the table.

Again hockey-stick curves are included, but each plot contains one curve per set of results
As with the simple report, the cbt_config.yaml files are included for the baseline results, and for any other test run where there is more than 20 lines of changes between it and the baseline yaml.

### Example comparison reports

[2_way.pdf](https://github.com/user-attachments/files/18380213/2_way.pdf)

[3-way.pdf](https://github.com/user-attachments/files/18380204/3-way.pdf)

For an example of a markdown report uploaded to github look at [this](https://github.com/lee-j-sanders/perfresults/blob/main/reports/7thJan_alexls12_4k_32k_64k_vs_main_4k/comparitive_performance_report_250107_131006.md) report.

## Testing

This has been extensively tested using the included scripts mentioned above. In addition the following pandoc versions have been tested:
pandoc 3.1.11.1 on fedora 41
pandoc 2.14.0.3 on centos 9 and RHEL 9.5


lee-j-sanders (approved):
I've used Chris' comparison tool several times upto 5 sets of data in parallel and it works brilliantly! No risk of regression to CBT as this is new function. Approved.

PR #324, opened 2025-01-30
Adding a total_iodepth option that can be used with librbdfio

harriscr:
## Description
Add an option to the CBT configuration yaml file called total_iodepth. The total_iodepth is then split evenly among the number of volumes used for the test. If the total_iodepth for a run does not divide evenly into the number of volumes, then any remainder will be assigned 1iodepth at a time to the volumes, starting at 0.

For (a simple) example: 
For an total_iodepth of 18 and volumes_per_client of 5, the following iodepth allocations would occur:
| volume | iodepth |
| :---: | :---:|
| 0 | 4 |
| 1 | 4 |
| 2 | 4 |
| 3 | 3 |
| 4 | 3 |

If the number of volumes specified is such that there is not enough iodepth for 1 per volume, then the number of volumes for that test will be reduced so that an iodepth of 1 per volume can be achieved.

Example:
For volumes_per_client = 5 and total_iodepth=4, the benchmark would be run with 4 volumes, each of iodepth 1

## Testing

Manual testing was done for a number of scenarios, with and without using workloads The below are the output from debug statements added for the purposes of testing.

### Regression testing:  iodepth=32, volumes_per_client=3
CHDEBUG: fio_cmd for volume 0 is /usr/local/bin/fio --ioengine=rbd --clientname=admin --pool=cbt-librbdfio --rbdname=cbt-librbdfio-`hostname -f`-0 --invalidate=0 --rw=randwrite --output-format=json,normal --runtime=300 --numjobs=1 --direct=1 --bs=4096B **--iodepth=32** --end_fsync=0 --norandommap --write_iops_log=/tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/op_size-00004096/concurrent_procs-003/iodepth-032/randwrite/output.0 --write_bw_log=/tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/op_size-00004096/concurrent_procs-003/iodepth-032/randwrite/output.0 --write_lat_log=/tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/op_size-00004096/concurrent_procs-003/iodepth-032/randwrite/output.0 --log_avg_msec=101 --name=cbt-librbdfio-`hostname -f`-file-0  > /tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/op_size-00004096/concurrent_procs-003/iodepth-032/randwrite/output.0

[...]

PR #326, opened 2025-02-12
Add pre_condtime for preconditioning and allow multiple randrw workloads within the same YAML

lee-j-sanders:
This PR only changes the functionality of workloads.

If workloads are in use, and there is a preconditioning step within the YAML, often users will want to run preconditioning for say 10 minutes, where as each other workload will run for a shorter period of time, this PR allows the user to override **time** with **precond_time** and specify **precond** within the preconditioning step to use the different time attribute.

Additionally, workloads with randrw only let you specify 1 mixed workload within a YAML as the same directory name was used for the subsequent randrw workloads, thus overwriting all the previous results in the archive directory with the later tests. This PR adds the readwrite ratio into the directory name if the mode is randrw.

The directory name for 100% read and 100% write tests are unaffected.

### DIFF
--- a/benchmark/librbdfio.py
+++ b/benchmark/librbdfio.py
@@ -29,6 +29,8 @@ class LibrbdFio(Benchmark):
         self.recov_test_type = config.get('recov_test_type', 'blocking')
         self.data_pool_profile = config.get('data_pool_profile', None)
         self.time = config.get('time', None)
+        self.precond_time = config.get('precond_time',None )
+
         # Global FIO options can be overwritten for specific workload options
         # would be nice to have them as a separate class -- future PR
         self.time_based = bool(config.get('time_based', False))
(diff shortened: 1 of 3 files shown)


lee-j-sanders:
Note, at the moment this is based on https://github.com/ceph/cbt/pull/324 - at the time of writing is unmerged.

benchmark/librbdfio.py
```diff
@@ -29,6 +29,8 @@ def __init__(self, archive_dir, cluster, config):
         self.recov_test_type = config.get('recov_test_type', 'blocking')
         self.data_pool_profile = config.get('data_pool_profile', None)
         self.time = config.get('time', None)
+        self.precond_time = config.get('precond_time',None )
```
perezjosibm, line 32: If you look at the way workloads work, basically they temporarily overwrite the global FIO options defined in the test plan .yaml (normally located *before* the workloads section). So within a precondition one can set a time value that only affects the precondition, which is then restored to the global. Then, another workload step would do the same.

In other words, there is not actual need to introduce yet another 'precond_time'!

But you can do the experiment: define two different time values in a precond and a workload step, from the global ones, with the existing code (that is, without this change) to convince yourself.

perezjosibm (changes requested):
I don't think the new option for precond_time is needed, I invite the submitter to try the suggested experiment without the change proposed to help convince himself.

benchmark/librbdfio.py
perezjosibm: And this is also mentioned in the documentation https://github.com/ceph/cbt/blob/master/docs/Workloads.md

harriscr:
The pytest warning doesn't stop the test from being run, so all run successfully. We are getting the warning because pytest assumes that anything with test in the title is actually a test case and tries to treat it as such. See [the default glob pattern](https://docs.pytest.org/en/stable/reference/reference.html#confval-python_classes)

Interestingly I don't see the warning when running the tests via vscode, only on the cli, which is something I want to investigate more.

There are 2 ways to remove the warning. One is to add `__test__ = False` to the TestRunResult class itself, and the other would be to set `TestRunResult.__test__ = False` in the test file (test_common_output_formatter.py)

The second option gives a mypy `attr-defined` error, so the first seems to be the best way to go.
Will discuss and either submit a PR to do this, or add the change in my next PR

perezjosibm:
I had a quick look: In the constructor, we load the workloads as a dict, if not empty we backup the global FIO options, which may include iodepth, etc -- any option that is used for mkfiocmd().

```python
        self.workloads = config.get('workloads', {})
        if self.workloads:
            self.backup_global_fio_options()
```
 **_Notice thought that it does not include 'time',_** (or more appropriate, 'runtime' per workload step), so to make the required runtime to be overloaded by each workload step, simply add the line (or a more appropriate default time) in the constructor -- and corresponding backup/restore class methods:
```python
 self.time = config.get('time', None)
```
Otherwise, adding yet another option explicitly ('precond_time') will clutter the code.

PR #328, opened 2025-02-24
Fixes for post_processing code for the total_iodepth feature

harriscr:
With the inclusion of the total_iodepth feature in [PR324](https://github.com/ceph/cbt/pull/324) the post_processing modules now give eratic results. 
This is because currently the code uses the iodepth value from the fio logs, and not the total_iodepth value set as that is never passed to fio.

The temporary fix is to parse the total_iodepth number from the file path used for the fio results run via CBT (taken from the write_iops_log value stored in the fio output file.

A better fix can be done when the workloads code is split into its own module, and we can write the total_iodepth value somewhere that the post_processing code can read, or we have a better way to cnnect the fio results to the workload that ran it. Until then, this code will allow the post processing to work with the total_iodepth feature

### DIFF
--- a/post_processing/formatter/test_run_result.py
+++ b/post_processing/formatter/test_run_result.py
@@ -86,7 +86,9 @@ class TestRunResult:
 
         with open(str(file_path), "r", encoding="utf8") as file:
             data: dict[str, Any] = json.load(file)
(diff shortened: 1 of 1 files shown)


lee-j-sanders (approved):
Looks good to me!

PR #329, opened 2025-02-25
Adding an option for a pre-workload script that is run before the IO exerciser

harriscr:
This PR adds the "pre_workload_script" to a workload specified in the configuration yaml file. It is specific to the workload, and has no effect outside of the workloads section of the yaml file.

For more information see the [workloads](https://github.com/ceph/cbt/blob/master/docs/Workloads.md) documentation in CBT.

It is the user responsibility to make sure that the script is available at the same path for all clients used in the test run. 
The script must exit with a return code that matches the OS schema i.e. for Linux return code 0 is good, and anything else is considered bad.

The script is run on every client using the standard common.pdsh() method in CBT.

This has been tested locally with a script that deletes and recreates volumes for a workload to run on before the fio IO is run. See the below output snippet gtom the cbt log file that shows the script being scheduled and run:

```
22:05:48 - DEBUG    - cbt      - Scheduling script /cbt.lee/tools/setup_cluster/mkdelvols.cbt to run
22:05:48 - DEBUG    - cbt      - CheckedPopen continue_if_error=True, shell=False args=pdsh -f 1 -R ssh -w root@HOST /cbt.lee/tools/setup_cluster/mkdelvols.cbt
22:06:10 - DEBUG    - cbt      - Using rbdname cbt-librbdfio-`hostname -f`-0
22:06:10 - DEBUG    - cbt      - CheckedPopen continue_if_error=True, shell=False args=pdsh -f 1 -R ssh -w root@HOST sudo /usr/local/bin/fio --ioengine=rbd --clientname=admin --pool=rbd_replicated --rbdname=cbt-librbdfio-`hostname -f`-0 --invalidate=0 --rw=write --output-format=json --runtime=60 --time_based --ramp_time=30 --numjobs=1 --direct=1 --bs=32768B --iodepth=1 --end_fsync=0 --norandommap --write_iops_log=/tmp/cbt/00000000/LibrbdFio/write_32768/iodepth-002/numjobs-001/output.0 --write_bw_log=/tmp/cbt/00000000/LibrbdFio/write_32768/iodepth-002/numjobs-001/output.0 --write_lat_log=/tmp/cbt/00000000/LibrbdFio/write_32768/iodepth-002/numjobs-001/output.0 --log_avg_msec=100 --name=cbt-librbdfio-`hostname -f`-file-0  > /tmp/cbt/00000000/LibrbdFio/write_32768/iodepth-002/numjobs-001/output.0
[...]

PR #333, opened 2025-05-19
New workload classes

harriscr:
This is the first part of the work to allow the workloads feature added in [PR 306](https://github.com/ceph/cbt/pull/306) to be used with any benchmark type.

To achieve this a number of new classes have been added to CBT. UML class digrams for these have been generated using pyreverse, and are given in each section below.

To integrate Workloads into any particular benchmark type a Workloads object would need to be added to the class variables, and instantiated by passing the configuration object used to create the benchmark.
A new command class for the benchmark type would also have to be created so that the workload can understand how to convert the yaml options to a CLI invocation that can be used to run the I/O exerciser in question.

The easiest way would be to add a self._workloads to the Benchmark base class and instantiate it there with the config object and archive directory.
```
self._workloads: Workloads = Workloads(config, self.run_dir)
```

Then the specific benchmark can run the workload using:
```
if self._workloads.exist():
            self._workloads.set_benchmark_type(<string_representation_of_benchmark_type>)
            self._workloads.set_executable(<full_path_to_executable_for_benchmark>)
            self._workloads.run()
```

### Class diagram
workload_classes

### Workloads
This class is a container for the actual workload classes themselves. It is instantiated with the workloads section of the benchmark yaml file, and is used to run the individual workloads.

### Workload
The Workload class contains the details for each individual workload. It is designed to be created and called via the Workloads class, so should never be called directly.

## Command classes
These classes try to encapsulate the CLI command that will eventually be run on the system to run the I/O exerciser. Eventually the aim is to add a command class for each individual I/O exerciser

### Class diagram
command_classes

### Command
An abstract base class for  any command type.

### FioCommand
The concrete class for an fio command. Parses all the options passed via a CBT yaml file for a single run of the fio I/O exerciser. This may need to be split farther in future into rbd and non-rbd versions in the future, but for this initial code a single class is sufficient

### CliOptions
A class based on the standard python dictionary that holds key/value pairs that equate to an option passed used on the CLI and the corresponding value for that option. Unlike a regular dictionary it does not allow values to be updated, and returns None instead of a KeyError when a value for an unknown key is requested.
Note to reviewers: I'm not sure if this class is needed or not, but it seems to be a neat way to cope with sets of options and corresponding values for a CLI invocation

## Testing
```
0.02s setup    tests/test_bm_cephtestrados.py::TestBenchmarkcephtestrados::test_valid_archive_dir
0.01s setup    tests/test_bm_rbdfio.py::TestBenchmarkrbdfio::test_valid_archive_dir
0.01s setup    tests/test_bm_kvmrbdfio.py::TestBenchmarkkvmrbdfio::test_valid_archive_dir
0.01s setup    tests/test_bm_librbdfio.py::TestBenchmarklibrbdfio::test_valid_archive_dir
0.01s setup    tests/test_bm_rawfio.py::TestBenchmarkrawfio::test_valid_archive_dir
======================== 326 passed, 3 skipped in 0.44s ========================
Finished running tests!
```
Black and ruff show no errors.

I'll update teuthology logs once they have run

### DIFF
--- a/cli_options.py
+++ b/cli_options.py
@@ -19,7 +19,7 @@ class CliOptions(UserDict[str, Optional[str]]):
         if key not in self.data.keys():
             self.data[key] = value
         else:
-            log.warning("Not adding %s:%s to configuration. A value is already set", key, value)
+            log.debug("Not adding %s:%s to configuration. A value is already set", key, value)
 
     def __update__(self, key_value_pair: tuple[str, str]) -> None:
         """
@@ -30,7 +30,7 @@ class CliOptions(UserDict[str, Optional[str]]):
         if key not in self.data.keys():
             self.data[key] = value
         else:
-            log.warning("Not Updating %s:%s in configuration. Value already exists", key, value)
+            log.debug("Not Updating %s:%s in configuration. Value already exists", key, value)
 
     def __getitem__(self, key: str) -> Optional[str]:
         """
@@ -40,7 +40,7 @@ class CliOptions(UserDict[str, Optional[str]]):
         if key in self.data.keys():
             return self.data[key]
         else:
-            log.warning("Key %s does not exist in configuration", key)
+            log.debug("Key %s does not exist in configuration", key)
             return None
 
     def clear(self) -> None:
--- a/cli_options.py
+++ b/cli_options.py
@@ -14,7 +14,7 @@ class CliOptions(UserDict[str, Optional[str]]):
     """
     Thic class encapsulates a set of CLI options that can be passed to a
     command line invocation. It is based on a python dictionary, but with
-    behaviour modified  so that duplicate entries do not update the original.
+    behaviour modified so that duplicate keys do not update the original.
     """
 
     def __setitem__(self, key: str, value: Optional[str]) -> None:
--- a/cli_options.py
+++ b/cli_options.py
@@ -1,0 +1,53 @@
+"""
+A class to encapsulate a set of configuration options that can be used to
+construct the CLI to use to run a benchmark
+"""
+
+from collections import UserDict
+from logging import Logger, getLogger
+from typing import Optional
+
+log: Logger = getLogger("cbt")
+
+
+class CliOptions(UserDict[str, Optional[str]]):
+    def __setitem__(self, key: str, value: Optional[str]) -> None:
+        """
+        Add an entry to the configuration.
+        Will report an error if key already exists
+        """
+        if key not in self.data.keys():
+            self.data[key] = value
+        else:
+            log.warning("Not adding %s:%s to configuration. A value is already set", key, value)
+
+    def __update__(self, key_value_pair: tuple[str, str]) -> None:
+        """
+        Update an existing entry in the configuration.
+        If the entry exists then don't update it
+        """
+        key, value = key_value_pair
+        if key not in self.data.keys():
+            self.data[key] = value
+        else:
+            log.warning("Not Updating %s:%s in configuration. Value already exists", key, value)
+
+    def __getitem__(self, key: str) -> Optional[str]:
+        """
+        Get the value for key in the configuration.
+        Return None and log a warning if the key does not exist
+        """
+        if key in self.data.keys():
+            return self.data[key]
+        else:
+            log.warning("Key %s does not exist in configuration", key)
+            return None
+
+    def clear(self) -> None:
+        """
+        Clear the configuration
+        """
+        self.data = {}
+
+    def __str__(self) -> str:
+        return super().__str__()
--- a/cli_options.py
+++ b/cli_options.py
@@ -11,6 +11,12 @@ log: Logger = getLogger("cbt")
 
 
 class CliOptions(UserDict[str, Optional[str]]):
+    """
+    Thic class encapsulates a set of CLI options that can be passed to a
+    command line invocation. It is based on a python dictionary, but with
+    behaviour modified  so that duplicate entries do not update the original.
+    """
+
     def __setitem__(self, key: str, value: Optional[str]) -> None:
         """
         Add an entry to the configuration.
@@ -48,6 +54,3 @@ class CliOptions(UserDict[str, Optional[str]]):
         Clear the configuration
         """
         self.data = {}
-
-    def __str__(self) -> str:
-        return super().__str__()
--- a/command/command.py
+++ b/command/command.py
@@ -73,7 +73,7 @@ class Command(metaclass=ABCMeta):
         """
         Return the output directory that will be used for this command
         """
-        return self._output_directory
+        return self._generate_output_directory_path()
 
     def set_executable(self, executable_path: str) -> None:
         """
--- a/command/command.py
+++ b/command/command.py
@@ -1,0 +1,99 @@
+"""
+A class to deal with a command that will run a single instance of
+a benchmark executable
+
+It will return the full executable string that can be used to run a
+cli command using whatever method the Benchmark chooses
+"""
+
+from abc import ABCMeta, abstractmethod
+from logging import Logger, getLogger
+from typing import Optional
+
+from cli_options import CliOptions
+
+log: Logger = getLogger("cbt")
+
+
+class Command(metaclass=ABCMeta):
+    """
+    A class that encapsulates a single CLI command that can be run on a
+    system
+    """
+
+    def __init__(self, options: dict[str, str]) -> None:
+        self._executable: Optional[str] = None
+        self._output_directory: str = ""
+        self._options: CliOptions = self._parse_options(options)
+
+    @abstractmethod
+    def _parse_options(self, options: dict[str, str]) -> CliOptions:
+        """
+        Take the options passed in from the configuration yaml file and
+        convert them to a list of key/value pairs that match the parameters
+        to pass to the benchmark executable
+        """
+
+    @abstractmethod
+    def _generate_full_command(self) -> str:
+        """
+        generate the full cli command that will be sent to the client
+        to run the benchmark
+        """
+
+    @abstractmethod
+    def _parse_global_options(self, options: dict[str, str]) -> CliOptions:
+        """
+        Parse the set of global options into the correct format for the command type
+        """
+
+    @abstractmethod
+    def _generate_output_directory_path(self) -> str:
+        """
+        Generate the part of the output directory that is relevant to this
+        specific command.
+
+        The format is dependent on the specific Command implementation
+        """
+
+    def get(self) -> str:
+        """
+        get the full cli string that can be sent to a system.
+
+        This string contains all the options for a single run of the
+        benchmark executable
+        """
+        if self._executable is None:
+            log.error("Executable has not yet been set for this command.")
+            return ""
+
+        return self._generate_full_command()
+
+    def get_output_directory(self) -> str:
+        """
+        Return the output directory that will be used for this command
+        """
+        return self._output_directory
+
+    def set_executable(self, executable_path: str) -> None:
+        """
+        set the executable to be used for this command
+        """
+        self._executable = executable_path
+
+    def set_global_options(self, global_options: dict[str, str]) -> None:
+        """
+        Update the global options
+        """
+        self._options.update(self._parse_global_options(global_options))
+
+    def update_options(self, new_options: dict[str, str]) -> None:
+        """
+        Update the command with the new_options dictionary
+        """
+        self._options.update(new_options)
+        for key, value in new_options.items():
+            if key not in self._options.keys():
+                self._options[key] = value
+            else:
+                log.debug("key %s already exists. Not overwriting", key)
--- a/command/command.py
+++ b/command/command.py
@@ -92,8 +92,3 @@ class Command(metaclass=ABCMeta):
         Update the command with the new_options dictionary
         """
         self._options.update(new_options)
-        for key, value in new_options.items():
-            if key not in self._options.keys():
-                self._options[key] = value
-            else:
-                log.debug("key %s already exists. Not overwriting", key)
--- a/command/fio_command.py
+++ b/command/fio_command.py
@@ -41,7 +41,6 @@ class FioCommand(Command):
 
         fio_cli_options.update(self.REQUIRED_OPTIONS)
         for option in self.DIRECT_TRANSLATIONS:
-            # Below only needed if testing?
             fio_cli_options[option] = options[option] if option in options.keys() else ""
 
         fio_cli_options["rw"] = options.get("mode", "write")
@@ -49,7 +48,7 @@ class FioCommand(Command):
         fio_cli_options["pool"] = options.get("poolname", "cbt-librbdfio")
         fio_cli_options["numjobs"] = options.get("numjobs", "1")
         fio_cli_options["bs"] = options.get("op_size", "4194304")
-        fio_cli_options["end_fsync"] = f"{options.get('end_fsync', 0)}"
+        fio_cli_options["end_fsync"] = f"{options.get('end_fsync', '0')}"
 
         if options.get("random_distribution", None) is not None:
             fio_cli_options["random_distribution"] = options.get("random_distribution", None)
--- a/command/fio_command.py
+++ b/command/fio_command.py
@@ -1,0 +1,178 @@
+"""
+A class to deal with a command that will run a single instance of the
+fio I/O exerciser
+
+It will return the full executable string that can be used to run a
+cli command using whatever method the Benchmark chooses
+"""
+
+from logging import Logger, getLogger
+from typing import Optional
+
+from cli_options import CliOptions
+from command.command import Command
+from common import get_fqdn_cmd
+
+log: Logger = getLogger("cbt")
+
+
+class FioCommand(Command):
+    """
+    The fio command class. This class represents a single fio command
+    line that can be run on a local or remote client system.
+    """
+
+    REQUIRED_OPTIONS = {"ioengine": "rbd", "clientname": "admin", "invalidate": "0", "direct": "1"}
+    DIRECT_TRANSLATIONS: list[str] = ["numjobs", "iodepth"]
+
+    def __init__(self, options: dict[str, str], workload_output_directory: str) -> None:
+        self._volume_number: int = int(options["volume_number"])
+        self._total_iodepth: Optional[str] = options.get("total_iodepth", None)
+        self._workload_output_directory: str = workload_output_directory
+        super().__init__(options)
+
+    def _parse_global_options(self, options: dict[str, str]) -> CliOptions:
+        global_options: CliOptions = CliOptions(options)
+
+        return global_options
+
+    def _parse_options(self, options: dict[str, str]) -> CliOptions:
+        fio_cli_options: CliOptions = CliOptions()
+
+        fio_cli_options.update(self.REQUIRED_OPTIONS)
+        for option in self.DIRECT_TRANSLATIONS:
+            # Below only needed if testing?
+            fio_cli_options[option] = options[option] if option in options.keys() else ""
+
+        fio_cli_options["rw"] = options.get("mode", "write")
+        fio_cli_options["output-format"] = options.get("fio_out_format", "json,normal")
+        fio_cli_options["pool"] = options.get("poolname", "cbt-librbdfio")
+        fio_cli_options["numjobs"] = options.get("numjobs", "1")
+        fio_cli_options["bs"] = options.get("op_size", "4194304")
+        fio_cli_options["end_fsync"] = f"{options.get('end_fsync', 0)}"
+
+        if options.get("random_distribution", None) is not None:
+            fio_cli_options["random_distribution"] = options.get("random_distribution", None)
+
+        if options.get("log_avg_msec", None) is not None:
+            fio_cli_options["log_avg_msec"] = options.get("log_avg_msec", None)
+
+        if options.get("time", None) is not None:
+            fio_cli_options["runtime"] = options.get("time", None)
+
+        if options.get("ramp", None) is not None:
+            fio_cli_options["ramp_time"] = options.get("ramp", None)
+
+        if options.get("rate_iops", None) is not None:
+            fio_cli_options["rate_iops"] = options.get("rate_iops", None)
+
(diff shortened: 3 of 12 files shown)


cli_options.py
```diff
@@ -0,0 +1,56 @@
...
+
+log: Logger = getLogger("cbt")
+
+
+class CliOptions(UserDict[str, Optional[str]]):
+    """
+    Thic class encapsulates a set of CLI options that can be passed to a
+    command line invocation. It is based on a python dictionary, but with
+    behaviour modified  so that duplicate entries do not update the original.
+    """
```
perezjosibm, line 18: I think it would be very helpful to illustrate with an example of the use of this feature. A simple unit test would do.

command/fio_command.py
```diff
@@ -0,0 +1,177 @@
+"""
+A class to deal with a command that will run a single instance of the
+fio I/O exerciser
+
```
perezjosibm, line 4: nit: s/fio/FIO/g

command/fio_command.py
```diff
@@ -0,0 +1,177 @@
...
+        # Secondary options
+        if fio_cli_options["rw"] == "readwrite" or fio_cli_options["rw"] == "randrw":
+            read_percent: str = options.get("rwmixread", "50")
+            write_percent: str = f"{100 - int(read_percent)}"
+            fio_cli_options["rwmixread"] = read_percent
+            fio_cli_options["rwmixwrite"] = write_percent
+
+        rbd_name: str = options.get("rbdname", "")
+        if rbd_name == "":
+            rbd_name = f"cbt-fio-`{get_fqdn_cmd()}`-{self._volume_number:d}"  # type: ignore[no-untyped-call]
```
perezjosibm, line 89: The expressions enclosed by backticks are evaluated via bash in the target machine (client endpoint), that's unfortunately the existing convention, not sure I like that. An alias is probably sufficient, shorter, and a small table (dict) collecting fqdn and aliases can be constructing during init?

command/fio_command.py
```diff
@@ -0,0 +1,177 @@
...
+            fio_cli_options["ramp_time"] = options.get("ramp", None)
+
+        if options.get("rate_iops", None) is not None:
+            fio_cli_options["rate_iops"] = options.get("rate_iops", None)
+
+        if bool(options.get("time_based", False)) is True:
+            fio_cli_options["time_based"] = ""
+
+        if bool(options.get("no_sudo", False)) is False:
+            fio_cli_options["sudo"] = ""
```
perezjosibm, line 72: This looks like an option not exclusive to FIO, so probably consider part of the parent class instead.
Also notice that FIO supports some other engines (like AIO) besides RBD that are useful for Ceph. These can be implemented as subclasses of FIO in this same module, only difference would be the set of options.

command/fio_command.py
```diff
@@ -0,0 +1,177 @@
...
+
+        return command
+
+    def _generate_output_directory_path(self) -> str:
+        """
+        For an fio command the output format is:
+        numjobs-<numjobs>/total_iodepth-<total_iodepth>/iodepth-<iodepth>
+        if total_iodepth was used in the options, otherwise:
+        numjobs-<numjobs>/iodepth-<iodepth>
+        """
```
perezjosibm, line 138: How difficult would it be if you need to track some other fields?

command/fio_command.py
```diff
@@ -0,0 +1,177 @@
...
+        Get the name for this job to give to fio
+        This is of the format:
+
+        cbt-<workload_name>-<hostname>-
+        """
+
+        job_name: str = ""
+
+        for process_number in range(processes_per_volume):
+            job_name += f"cbt-fio-{parent_workload_name}-`hostname`-file-{process_number} "
```
perezjosibm, line 159: Do we need to be consistent with the above convention of using the fqdn? The alias table suggestion would make things easier in the long term.

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def get_script_command(self) -> Optional[str]:
+        """
+        If the yaml specifies a script to be run before this workload then
+        return the command line invocation, otherwise return None
+        """
+        if self._script == "":
+            return None
+
+        return self._script
```
perezjosibm, line 90: `return self._script or None`

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+        self._executable_path: str
+        self._script: str = f"{options.get('pre_workload_script', '')}"
+
+    def get_commands(self) -> Generator[str, None, None]:
+        """
+        Return all I/O exerciser commands that need to be run to fully execute
+        this workload
+        """
+        if not self._commands:
+            self._create_commands_from_options()
```
perezjosibm, line 43: Why cannot be created during __init__ above?

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+"""
+
+from logging import Logger, getLogger
+from typing import Generator, Optional, Union
+
+from benchmarkfactory import all_configs  # pyright: ignore[reportUnknownVariableType]
+from command.command import Command
+from command.fio_command import FioCommand
+
+WORKLOAD_TYPE = dict[str, Union[str, list[str]]]  # pylint: disable=["invalid-name"]
```
perezjosibm, line 13: We probably need WORKLOAD_TYPE to be a class itself, because it encapsulates useful info, like number of OSD of the cluster, whether its Classic or Crimson, whether we need a number of Reactors, Alien threads, etc.
Notice also that can be part of the info you need for post-processing.

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        for key, value in global_options.items():
+            if key not in self._all_options.keys():
+                self._all_options[key] = value
+
+    def _create_command_class(self, options: dict[str, str]) -> Command:
+        """
+        Create the concrete command classes for each command for this workload
+        """
+        if self._parent_benchmark_type == "fio":
```
perezjosibm, line 121: Another nit: the current FIO class above is actually FIO RBD, if/when CBT supports another FIO engine (AIO, etc) From the list by 
`# fio --enghelp`
at least three engines are relevant for Ceph.

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __init__(self, name: str, options: WORKLOAD_TYPE, base_run_directory: str) -> None:
+        self._name: str = name
+        self._base_run_directory: str = base_run_directory
+        self._commands: list[Command] = []
+        self._parent_benchmark_type: Optional[str] = None
+        self._all_options: WORKLOAD_TYPE = options.copy()
+        self._executable_path: str
+        self._script: str = f"{options.get('pre_workload_script', '')}"
+
```
perezjosibm, line 36: As a minimum, a workload can be specified (regardless of benchmark) by the following parameters:

- IO type, 
- Block size, 
- IO depth, 
- (IO size normally the full target), 
- target (device or volume)
Of course the free dict _all_options, supports that, but if we have this in place already would be very useful for consistency across the code base, including post processing. This can be serialised in to a .json object and load it when post-processing.

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+            self._create_commands_from_options()
+
+        for command in self._commands:
+            yield command.get_output_directory()
+
+    def get_name(self) -> str:
+        """
+        Return the name of this workload
+        """
+        return self._name
```
perezjosibm, line 74: Some methods to quickly extract the details of the workload (IO type, IO size, etc) from its name would be very useful.

workloads/workloads.py
```diff
@@ -0,0 +1,141 @@
...
+        for workload_name, workload_options in workload_json.items():
+            workload = Workload(workload_name, workload_options, self._base_run_directory)
+            workload.add_global_options(self._global_options)
+            # workload.set_benchmark_type(self._benchmark_type)
+
+            self._workloads.append(workload)
+
+    def _get_global_options_from_configuration(self, configuration: BENCHMARK_CONFIGURATION_TYPE) -> WORKLOAD_TYPE:
+        """
+        Get any configuration options from the yaml that are not workload
```
perezjosibm, line 125: Suggestion: use the terminology to be consistent with the documentation:
.. from the Test Plan .yaml ...

workloads/workloads.py
```diff
@@ -0,0 +1,141 @@
...
+        specific
+        """
+        global_options: WORKLOAD_TYPE = {}
+
+        for option_name, value in configuration.items():
+            if option_name == "workloads" or option_name == "prefill":
+                # prefill is not an option for workloads as it is used in the Benchmark prefill()
+                # method.
+                # Workloads we also want to ignore
+                pass
```
perezjosibm, line 135: Can you please show in an example, or a unit test what is the expected layout of this new section in the test plan .yaml?

workloads/workloads.py
```diff
@@ -0,0 +1,141 @@
...
+            elif isinstance(value, list):
+                global_options[option_name] = value
+            else:
+                global_options[option_name] = f"{value}"
```
perezjosibm, line 139: If you convert single/scalar values to  strings, why is not requited for  list of items?

workloads/workloads.py
```diff
@@ -0,0 +1,141 @@
...
+
+        return global_options
```
perezjosibm, line 141: Would this PR also provide the code for calling this class from the main execution?

perezjosibm (changes requested):
Leaving review comments as per requested review, eventhough the state of the pr is still Draft.

command/fio_command.py
harriscr: Done

command/fio_command.py
harriscr: I agree - I don't like it either. It can introduce dependencies on the Operating System on the target machine which could be problematic. We should at least try and use a standard Python method for getting this in the future

My prefered solution would be to have the FQDN and alias stored as a class variable with a get() method that can be called to get the value, probably in the Cluster object. There is an argument that there should be a new Host, or System object to keep these details as they aren't exactly a property of a cluster.
Something good to think about for the future.

command/fio_command.py
harriscr: Yes, this class is really for RBD fio at the moment, as that is currently the only benchmark class that supports workloads. I was planning on splitting the classes later on as we add workload support to the other FIO classes.

As I said in the PR "This may need to be split farther in future into rbd and non-rbd versions, ". which will be sub-classes of an FioCommand as you suggest, although I would lean toward putting them in separate files rather than having multiple classes in a single file.

command/fio_command.py
```diff
@@ -0,0 +1,177 @@
...
+            fio_cli_options["log_lat"] = ""
+
+        processes_per_volume: int = int(options.get("procs_per_volume", 1))
+
+        fio_cli_options["name"] = self._get_job_name(options["name"], processes_per_volume)
+
+        return fio_cli_options
+
+    def _generate_full_command(self) -> str:
+        command: str = ""
```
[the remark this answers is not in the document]
harriscr: Yes, I was thinking I needed to add unit tests for these new classes.

Will add some in.

command/command.py
```diff
@@ -0,0 +1,99 @@
...
+    def update_options(self, new_options: dict[str, str]) -> None:
+        """
+        Update the command with the new_options dictionary
+        """
+        self._options.update(new_options)
+        for key, value in new_options.items():
+            if key not in self._options.keys():
+                self._options[key] = value
+            else:
+                log.debug("key %s already exists. Not overwriting", key)
```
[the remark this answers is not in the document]
harriscr: Good spot - this is extra code left over from before the CliOptions class was created. I will remove it.
I prefer to keep a single class per file if possible, rather than multiple classes, so I will keep the current structure.

It would be nice to use the Command class to encapsulate any command that needs to be run in the future. This will likely need some refactoring of, or moving the methods in this base class as required

workloads/workloads.py
harriscr: Done

workloads/workloads.py
harriscr: There is no new section in the test plan .yaml. The format remains the same as today, with the workload being defined per benchmark type.

workloads/workloads.py
harriscr: No, that will be covered in a different PR. This is simply adding the class infrastructure required to supprt workloads.

I have included a suggestion on how to run Workloads in the text of the PR, and I will be updating the documentation as well.

workloads/workloads.py
```diff
@@ -0,0 +1,141 @@
...
+        set the benchmark type that will be used to run the workload
+
+        This must be done by the benchmark before it attempts to run
+        any commands
+        """
+        self._benchmark_type = benchmark_type
+
+    def set_executable(self, executable_path: str) -> None:
+        """
+        Set the executable to be used for the
```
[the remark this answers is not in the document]
harriscr: Will fix

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+        be the list of total_iodepth values if the total_iodepth key exists,
+        or the iodepth value if it does not
+        """
+        iodepth_key: str = "iodepth"
+        if "total_iodepth" in configuration_keys:
+            iodepth_key = "total_iodepth"
+
+        return iodepth_key
+
+    def _calculate_iodepth_per_volume(self, number_of_volumes: int, iodepth: int, iodepth_key: str) -> dict[int, int]:
```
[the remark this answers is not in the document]
harriscr: That is an interesting thought - I shall change the method names as you are correct, target rather than volume makes much more sense.

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+        Create the concrete command classes for each command for this workload
+        """
+        if self._parent_benchmark_type == "fio":
+            return FioCommand(options, f"{self._base_run_directory}/{self._name}")
+
+        log.error("Benchmark Class %s is not supported by workloads yet", self._parent_benchmark_type)
+        raise NotImplementedError
+
+    def _create_commands_from_options(self) -> None:
+        unique_options: dict[str, str]
```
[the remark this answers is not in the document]
harriscr: It cannot be called from the constructor as it requires the exectable and benchmark type to be set, and these are unknowns at the time the class is created.

workloads/workload.py
harriscr: Yes, I am aware. 
My thought was to make this generic to start with. As support for workloads is added to other benchmark types, including other flavours of FIO, the value here can be made more specific as required. This way we don't end up with a profusion of options with overlapping uses as we have in the benchmark classes today.
But maybe you are right, it would be better to be clear now in the code.

It would be nice to have a single fio class that would handle the options for any of the engine types, but in reality I think this would make the code far too messy and harder to maintain.

workloads/workload.py
harriscr: At the point of init() we do not have the executable to be used in the command or the benchmark type for this workload.
The benchmark type is required to be able to instantiate the correct concrete Command class, and the executable is required to construct the CLI to be run.

We cannot pass these to the init() as parameters because they are not known at the time the Workload is first created

workloads/workload.py
harriscr: The problem with adding methods for pullig out details from the name is that there is not a set format for the workload name. It can be any text string the creator of the test plan .yaml desires, and does not need to contain any details about the IO type and size.
This makes the task of parsing the name for these details very difficuly without first mandating a format for the workload name.

workloads/workload.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def get_name(self) -> str:
+        """
+        Return the name of this workload
+        """
+        return self._name
+
+    def has_script(self) -> bool:
+        """
+        True if there is a script for this workload, otherwise false
```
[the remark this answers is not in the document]
harriscr: The script option was added in https://github.com/ceph/cbt/pull/329 to allow a user-defined script to be run between each workload.
The particular use case the option was added for initially was to collect performance data for writing to an empty volume, as opposed to appending or over-writing data. This was achieved by deleting and re-creating every volume between each individual FIO run in a script.

When running the workload we need to know of a script is required to be run before a workload, and run it (see the run() method on workloads)

I'll update the comment in the code to make it clearer though

command/fio_command.py
harriscr: Not at all I think. They could be added in this method if required, although if there are too meany then I think a refactor would be required

The idea behind this was to encode as few details as possible in the file path.
The rest of the details for a run are included in the output json file for FIO, and as such are better read from there as there is less scope for the value used by FIO and the value in the path to be different.

If we discover things that are missing for other benchmark types then those can be added when workload support is added to that benchmark type

command/fio_command.py
harriscr: In this case I don't think so, no.
The job name isn't used for anything other than a section heading in the FIO output JSON file, and I don't think using the FQDN is necessary. I am not aware of anywhere that tries to match the RBD name with the run name, so we should be OK from that point of view
I dont think the job name is used at all by CBT, although I haven't thoroughly checked teuthology.
The RBD name, on the other hand, is fixed by the user if using an existing config, or created by CBT using the FQDN format which is also used above

As I mentioned in a previous comment, I would much prefer to have the FQDN and hostname as properties of a Cluster or System object as to me that is where they logically belong. Ce can use get methode to pull in the details here

workloads/workloads.py
harriscr: We do not want to convert the entire list to a string as that doesn't quite do what's expected.
e.g (using python 3.9)
>>> l: list[int] = [1, 2, 3, 6, 89]
>>> print(f"{l}")
[1, 2, 3, 6, 89]
>>> b = f"{l}"
>>> print(b[:-3])
[1, 2, 3, 6,

So the whole list, including the brackets would be converted to a single string, which would then have to be parsed at a later date and the brackets stripped off.
If we leave this list as a list, then we can iterate over it in later code without having to fist strip the brackets and then split the string into its component parts.
There is an argument to say we should do the same with dictionaries, but I haven't come across one yet - maybe when I'm doing more testing something will show up.

We could iterate over the contents of the list and also convert those to a string, but with the random nested structure present in the test plan yaml files we will never be able to convert everything correctly.

workloads/workload.py
harriscr: The WORKLOAD_TYPE here is a type hint to MyPy for checking the code. As we collect more of these I think we should look into creating a cbt_types.py file to contain them all.

I agree that storing things like the number of ODS etc. that you mentioned would be useful. To me that sort of intofmation is a property of the cluster under test, not a property of an I/O workload though. It would belong to the Cluster object.

I am wondering if eventually we want to have something like a results object that stores the configuration and options/results for a single benchmark run. I know we output the cluster configuration at the start of a run (in the benchmark .yaml file),
but that leave a step of matching the config to the benchmark run. For a 1:1 mapping this isn't too difficult, but if CBT ran with multiple cluster definitions then it makes things harder.
There is also already a Result object in the code (see benchmark.py), but that currently doesn't store any information about the configuration.

perezjosibm: The Results object class already exists, albeit it looks an initial try and needs TLC, have a look at `benchmark.py:`

```python
class Result:
    def __init__(self, run, alias, result, baseline, stmt, accepted):
        self.run = run
        self.alias = alias
        self.result = result
        self.baseline = baseline
        self.stmt = stmt
        self.accepted = accepted

    def __str__(self):
        fmt = '{run}: {alias}: {stmt}:: {result}/{baseline}  => {status}'
        return fmt.format(run=self.run, alias=self.alias, stmt=self.stmt,
                          result=self.result, baseline=self.baseline,
                          status="accepted" if self.accepted else "rejected")
```
I think it would be very useful to have a relation between  a 'run' with its corresponding (Cluster) configuration.

(Sorry I skipped the gun, just read the rest of the post and saying the same thing).

perezjosibm: > For a 1:1 mapping this isn't too difficult, but if CBT ran with multiple cluster definitions then it makes things harder.
I see your point, for example, for Crimson, I normally need a range over number of OSD (implicitly backend drives), number of reactors, number of Alien threads, etc. so diverge completely from the current test plan schema that CBT expects. that's why I capture succint details of the configuration parameters in the test run name itself, so each test run ends up with its own .json linking the configuration details.
I might not lobby for such a convention to be supported in CBT (since recreating clusters etc might conflict with the expected behaviour wrt teuthology, which I think runs CBT with flag indicating to use the existing cluster iirc). I might keep a prototype in my own checkout and test it in anger.

command/fio_command.py
perezjosibm: IIRC there were already separate Python modules per FIO class according to engine, but I might be wrong.

To make things better for the future, I'd suggest to have a minimal hierarchy in place before hand, migth save lots of trouble later on.

Also, it seems FIO might be used to support eg NVME (albeit with the Linux native AIO, but again, not 100% sure).

command/fio_command.py
perezjosibm: > Something good to think about for the future.

Quoting to Yoda:
_Hard to see the future is_

perezjosibm: If Commands are going to be used as generic for both the Cluster entities as well as (endpoint) hosts, a thin wrapper around a command with the target (as a key to a dictionary of the alias: hostname (or fqdn as appropriate) would be worth considering.

Notice that during monitoring, a (normally separate) connection to the target (eg Cluster entity or client endpoint) is keep since we run the monitoring process (eg. top, ceph tell, collectl) in the background. So keep a list of {Command: entity} as a collection/iterable under the target entity object (eg. Cluster with a list of open connections, same for each test client endpoint). This approach would benefit Restarting/retrying/killing as appropriate.

workloads/workload.py
perezjosibm: Yes, I mean as instance object attributes in the Workload class (self._x variables) 👍

command/fio_command.py
harriscr: I think it would be better for each Cluster (or endpoint?) object to have a command runner class that knows how to run a command on that endpoint. Then you can pass a Command to to the command runner and it will run it on it's owning end point.
That way the Command can be completely independent from the endpoint/Cluster.
There will be endpoint specific values that need set (rbd_name, possibly jobname), but they can be done before the command is actually run.

PR #337, opened 2025-07-18
Updating post processing

harriscr:
# Description
This PR covers two main themes:

1. Updating librbdfio
    - so that it runs the new worload code
    - so that it converts the fio output correctly into json format in the results directory
2. Improvement to the post-processing scripts from #319, #320 and #321
    - to be able to convert the output from running the new workloads code in #333 
    - generate_performance_report.py - so that it generates the intermediate file format for a run before creating the report, if required
    - generate_comparison_performance_report.py - so that it generates the intermediate file format for a run before creating the report, if required
    - better logging and a separate log file for post processing to aid debugging problems

# Testing

## Manual Testing
1. running the new workloads code
2. post processing the results
3. creating a report from the test run
4. creating a report from data from an old test run before #333 

## Teuthology
Results will be posted here once the tests complete
- perf-basic
- rados/perf
- crimson-rados perf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 *.pyc
 *.pyo
+*.venv
+*.code-workspace
+.devcontainer
+*.toml
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -8,6 +8,8 @@ import yaml
 import settings
 import common
 
+from workloads.workloads import Workloads
+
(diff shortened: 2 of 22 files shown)


lee-j-sanders (approved):
LGTM, I've been testing the ch_wip_workload_postprocessing branch with FIO and post processing results and it appears to be working as well as earlier prototype commits.

PR #339, opened 2025-08-29
Improvements to the report generation

harriscr:
A few quality of life improvements to the report generation added in #320 #321 and #322.

1. Move the plot legends outside of the graph axes so that it doesn't cover the data. 
         The automatic placement was good in most cases, but there were odd cases where data was obscured
2. Some plot titles were given to 1 decimal place - change this back to 0 decimal places
        e.g. **1024K** instead of **1024.0K**
3. An option to not draw error bars on the plots in a report.
        It was found that some error bars were so big that they ended up distorting the axes and stretching the plot
4. An option to force the regeneration of the intermediate data and plots from the raw data

In 3, and 4 above the current behaviour will remain the default. For 1 and 2 this will become the new default as it makes the charts look better.

### DIFF
--- a/tools/generate_comparison_performance_report.py
+++ b/tools/generate_comparison_performance_report.py
@@ -163,7 +163,7 @@ def main() -> int:
         "--force_refresh",
         action="store_true",
         required=False,
-        help="Regenerate the inetrmediate files and plots, even if they exist",
+        help="Regenerate the intermediate files and plots, even if they exist",
     )
 
     arguments: Namespace = parser.parse_args()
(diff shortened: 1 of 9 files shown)


tools/generate_comparison_performance_report.py
```diff
@@ -155,6 +159,13 @@ def main() -> int:
         help="The filename root of all the CBT output json files",
     )
 
+    parser.add_argument(
+        "--force_refresh",
+        action="store_true",
+        required=False,
+        help="Regenerate the inetrmediate files and plots, even if they exist",
```
lee-j-sanders, line 166: slight typo here

lee-j-sanders (approved):
A couple of very minor typos, otherwise looks good to me. Thanks!

PR #340, opened 2025-09-15
Automatic report generation: common report modules

harriscr:
Currently the report generation scripts:
- generate_performance_report.py
 - generate_comparison_performance_report.py 

use methods within the script to generate a report. This change moves away from that and uses a common module to generate the report. 

This is the first step to generating reports as part of a CBT run, if the correct options are set in the yaml file.

### Testing
Manually generated a single performance report and a comparison performance report using the new code. Verified that these had the same content as ones generated using the previous method

### DIFF
--- a/example/example-report_generation.yaml
+++ b/example/example-report_generation.yaml
@@ -1,0 +1,89 @@
+cluster:
+  user: 'perf'
+  head: "incerta01"
+  clients: ["incerta01"]
+  osds: ["incerta02"]
+  mons:
+    incerta01:
+      a: "10.0.10.101:6789"
+  mgrs:
+    incerta01:
+      a: ~
+  mdss:
+    incerta01:
+      a: ~
+  osds_per_node: 1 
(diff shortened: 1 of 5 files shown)


example/example-report_generation.yaml
```diff
@@ -0,0 +1,89 @@
...
+    iodepth: [32]
+    osd_ra: [4096]
+    cmd_path: '/home/perf/src/fio/fio'
+    log_avg_msec: 100
+
+report:
+    force_refresh: True
+    create_pdf: False
+    output_directory: "/home/perf/reports/"
+    results_file_root: "output"
```
lee-j-sanders, line 87: I'm wondering if this example should also had how to generate automatic comparisons,
eg. referencing the archives and the baseline.

Also mention no_error_bars on standalone reports.

lee-j-sanders (approved):
One comment on how to use comparisons in the example YAML. Other than that, looks good!

PR #347, opened 2025-12-02
Post Processing: Refactor of result modules

harriscr:
In order to be able to include resource usage statistics in the reports produced as part of the post processing some re-factoring of the post_processing/run_result modules was necessary.

This is mainly the creating of a run_result directory and creation of some base classes that can be extended for particular run types.

This is the first step in adding resource monitor result file parsing and adding to the common format output file (which we use to generate the reports). Future PRs will deal with:
 - parsing the various resource monitoring options that CBT supports, top, collectl, perf etc. into the common format.
 - creating resource monitoring plots for the report e.g. CPU and Memory utilisation over time
 - Additional report options to control which resource statistics are addre to a report

#### Testing
Teuthology perf-basic: https://pulpito.ceph.com/harriscr-2025-12-01_13:29:31-perf-basic-main-distro-default-smithi/
rados/perf: https://pulpito.ceph.com/harriscr-2025-12-02_08:34:46-rados:perf-main-distro-default-smithi/

All tests passed

### DIFF
--- a/benchmark/librbdfio.py
+++ b/benchmark/librbdfio.py
@@ -186,7 +186,8 @@ class LibrbdFio(Benchmark):
                 create_pdf = report_config.get("create_pdf", False),
                 force_refresh = report_config.get("force_refresh", False),
                 no_error_bars = report_config.get("no_error_bars", False),
-                comparison = False
+                comparison = False,
+                plot_resources = report_config.get("plot_resource", False)
             )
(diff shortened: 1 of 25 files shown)


lee-j-sanders (approved):
I've been using the comparison tool for the past couple of days without any issues. The change itself is quite large to review, but given I've been using this for a few days without problems, it should be good to merge.

PR #357, opened 2026-07-24
Add more resource plotting to the reports

harriscr:
The post processing modules in CBT can produce a plot with CPU usage during a run. Currently only the CPU usage as reported by the FIO process can be plotted.

CBT supports additional tools for collecting resource usage statistics such as top, collectl and perf.

This PR adds support for resource usage as reported by the top and collectl tools.

As part of this work a minor refactor of the monitoring classes was done to bring them up to modern python coding standards. Additionally the top monitoring relied on an OSD pid file being present, which is a very specific use case.
This is still supported, but as a sub-class of a more generic top monitoring.

IBM Bob 2.0 and 2.0.1 was used to assist with this code

*Testing*
This was tested by:
Creating a report containing top and collectl CPU usage [performance_report_260724_125457.pdf](https://github.com/user-attachments/files/30348475/performance_report_260724_125457.pdf)
A regression test for creating a report without and CPU uasage statistics.


harriscr:
Updated the CPU usage statistics from FIO to sum across processes, to give a total FIO CUP usage for a system

lee-j-sanders (approved):
Tested the latest bunch of commits. Generating nice CPU utilisation reports with collectl and top (measuring OSDs).

There is a niggly issue with needing to create a pids file for the OSDs, but we are working to resolve that in a separate PR.

Looks great! Thanks Chris

PR #364, opened 2026-09-03
logging: log to a single file by default

harriscr:
Currently CBT will log all messages to the CLI where it was called from by default. A log-file can be specified in the YAML, but this is not ideal, and many people tee or redirect the output from the CLI to a log file themselves.
It would be better if CBT logged everything to a log file, and only output messages that are required to show progress is happening to the CLI.

By default log cbt messages of *debug* to a log file and only show *info* and above on the screen. This will allow the CLI to show progress, without getting overwhelmed with many log messages. For debugging what went wrong the user can read the whole log file. 
Where appropriate additional log messages have been added to show progress on the CLI.
The log will be written to the specified archive directory to keep all the CBT output artefacts in a single location

Additionally this removes the post_processing logging and rolls it up into a single cbt-wide log file and configuration.

### Testing

Example CLI output:
```
INFO     - cbt: Workload 'seq32kread': parameter set 19/22 complete.
INFO     - cbt: Workload 'seq32kread': running parameter set 20/22 -> /tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/seq32kread/rbdfio/numjobs-002/iodepth-000096
INFO     - cbt: Ramp time: waiting 30s before collecting stats...
INFO     - cbt: Starting monitoring in /tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/seq32kread/rbdfio/numjobs-002/iodepth-000096
INFO     - cbt: Top monitoring collecting 30 samples (estimated ~90s)...
INFO     - cbt: Top monitoring collection complete.
INFO     - cbt: Stopping monitoring.
INFO     - cbt: Workload 'seq32kread': parameter set 20/22 complete.
INFO     - cbt: Workload 'seq32kread': running parameter set 21/22 -> /tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/seq32kread/rbdfio/numjobs-002/iodepth-000128
INFO     - cbt: Ramp time: waiting 30s before collecting stats...
INFO     - cbt: Starting monitoring in /tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/seq32kread/rbdfio/numjobs-002/iodepth-000128
INFO     - cbt: Top monitoring collecting 30 samples (estimated ~90s)...
INFO     - cbt: Top monitoring collection complete.
INFO     - cbt: Stopping monitoring.
INFO     - cbt: Workload 'seq32kread': parameter set 21/22 complete.
[...]

PR #365, opened 2026-09-15
Add progress bars to the CLI output

harriscr:
With PR #364 the cli output to the screen was reduced to show progress through the code, but remove all the extra debug statements to a log file. There is still an improvement that can be made though, showing how far through a particular CBT run we are as this is not obvious.

To do this we can add progress bars to the CLI output to give the user a better idea of how long their run may take, and how far through it they are. The bars are not shown when run through the teuthology suite as this would make the output less readable.

IBM Bob 2.0.1 and 2.0.3 were used to help with the change

###Example output###
```
INFO     - cbt: Starting monitoring in /tmp/cbt/00000000/LibrbdFio/osd_ra-00004096/seq32kwrite/rbdfio/numjobs-002/iodepth-000256                                                             
INFO     - cbt: Top monitoring collecting 30 samples (estimated ~90s)...                                                                                                                     
Overall run:  46%|███████████████████████████████████████████████████████████▍                                                                    | 2715.0/5850s  [elapsed 50:39 | eta 53:16]
Workload 'seq32kwrite' 18/20:  38%|██████████████████████████████████████████▊                                                                       | 45.0/120s  [elapsed 00:45 | eta 01:15]
```

###Testing###
All existing unit tests pass, and additional unit tests have been added for any new classes created.

```
0.08s call     tests/test_directory_comparison_plotter.py::TestDirectoryComparisonPlotter::test_draw_and_save_adds_legend
0.05s call     tests/test_time_series_plotter.py::TestTimeSeriesPlotter::test_plot_iops_with_zero_values
0.04s call     tests/test_monitoring_factory.py::test_get_object_returns_osd_top_monitoring
0.01s call     tests/test_time_series_formatter.py::TestFIOTimeSeriesParser::test_format_with_all_metrics
0.01s call     tests/test_fio_log_parser.py::TestFIOLogParser::test_parse_and_combine_logs_bw
================= 1012 passed, 3 skipped, 24 warnings in 1.45s =================
Finished running tests!
```

Black, ruff, mypy and pylint all pass


progress.py
```diff
@@ -0,0 +1,311 @@
...
+"""
+
+import contextlib
+import logging
+import sys
+import threading
+from collections.abc import Generator
+from typing import Any, Optional
+
+import tqdm
```
mohant-ibm: Might need to add `tqdm` and `typing-extensions` to requirements.txt.

cbt.py
```diff
@@ -41,6 +50,31 @@ def parse_args(args):
...
+        rebuild_every_test:   Whether the cluster is rebuilt for each benchmark.
+
+    Returns:
+        Estimated run time in seconds.
+    """
+    cluster_init_secs = progress.cluster_init_estimate()
+    total = cluster_init_secs  # at least one cluster init
+
+    for b in benchmarks_list:
+        if rebuild_every_test:
```
mohant-ibm: Would this count one extra `cluster_init_secs` when `rebuild_every_test = True`?

cbt.py
```diff
@@ -41,6 +50,31 @@ def parse_args(args):
...
+            total += cluster_init_secs
```
mohant-ibm: Would this count one extra `cluster_init_secs` when `rebuild_every_test = True`?

harriscr: Yes, we're adding an extra rebuild time. This is to handle the case when rebuild_every_test is False, but then the True logic is off by 1. tbh overestimating the run time is probably better than underestimating but that doesn't mean it shouldn't be fixed

progress.py
harriscr: Really thought I had, but it turns out not. Fixed in latest push

mohant-ibm (approved):
Looks good, thanks Chris.

PR #359, opened 2026-08-07
add Elbencho S3 benchmark with async parallel SSH

gitkenan:
## Summary

- Add a stdlib-only parallel SSH layer to `common.py` using `asyncio` + the system `ssh` binary, so benchmarks can opt out of `pdsh` with no new dependencies
- Introduce `benchmark/elbencho.py` — an S3 benchmark that runs its entire lifecycle (binary check, cleandir, dropcaches, workload fan-out, result pull) through the async SSH path
- Register Elbencho in `benchmarkfactory.py` (import + dict entry only — no changes to `get_all()` or `all_configs()`)
- 49 unit tests covering construction, validation, factory integration, CLI command building, run-loop fan-out, and pdsh-free lifecycle
- User-facing docs and a working example YAML config

### How Elbencho avoids Cartesian expansion

Elbencho's list-valued parameters (`threads`, `iodepth`, `blocksize`) live inside the `workloads:` block, not at the top level. `all_configs()` only sees top-level keys, which are all scalars or dicts, so it naturally yields a single config — no special-case code in the factory needed.

## Test plan

- [x] `python -m unittest tests.test_bm_elbencho` — 49 tests pass
- [x] Verify no regressions in existing benchmark tests

798 tests ran, all Elbencho tests (49) pass. The 7 failures are pre-existing post-processing test issues unrelated to our code.

- [x] End-to-end run against an RGW cluster with the example YAML

PR #366, opened 2026-09-15
Fix a type mismatch in 

harriscr:
As part of a code review for #359 it was noticed that in some instances the command classes were being initalised with a dict containing str, int, bool values when the  constructor was explicitly expecting strings. Fix that type mis-match so that the options passed in to the *command classes are definitely in a dict[str, Union[str, list[str]] format.

The code worked as written, but the type mismatch could lead to unexpected behaviour

IBM Bob 2.0.3 was used to help with this code.

###Testing###
New unit tests were written to catch this case.
All unit tests pass:
```
0.07s call     tests/test_directory_comparison_plotter.py::TestDirectoryComparisonPlotter::test_draw_and_save_adds_legend
0.06s call     tests/test_time_series_plotter.py::TestTimeSeriesPlotter::test_plot_time_series_creates_all_plots
0.05s call     tests/test_monitoring_perf.py::test_perf_start_remote_node_uses_pdsh
0.01s call     tests/test_time_series_formatter.py::TestFIOTimeSeriesParser::test_format_with_all_metrics
0.01s call     tests/test_fio_log_parser.py::TestFIOLogParser::test_parse_and_combine_logs_bw
================= 985 passed, 3 skipped, 21 warnings in 1.41s ==================
Finished running tests!
```

Manual testing shows the code working as expected after this change


workloads/workloads.py
```diff
@@ -154,12 +154,29 @@ def _create_configurations(self, workload_json: WorkloadYamlType) -> None:
...
+        """
+        Convert every option value (and every element within list values)
+        to a string so that dict[str, str | list[str]] is consistently
+        typed before being handed to Workload / FioCommand (which expect
+        dict[str, str]).
+        """
+        normalised: WorkloadType = {}
+        for key, value in options.items():
+            if isinstance(value, list):
+                normalised[key] = [f"{item}" for item in value]
```
gitkenan: In Python:

```bool(False)     # is False
bool("False")   # is True! Non-empty strings are always truthy.
```

Bob's recommendation: Either preserve bool types, convert booleans properly, or normalize boolean string handling in _parse_options() (e.g. str(value).lower() in ("true", "1")). Unconditionally formatting bool to f"{value}" breaks boolean semantics.

workloads/workloads.py
```diff
@@ -154,12 +154,29 @@ def _create_configurations(self, workload_json: WorkloadYamlType) -> None:
...
+    def _normalise_options(options: WorkloadType) -> WorkloadType:
+        """
+        Convert every option value (and every element within list values)
+        to a string so that dict[str, str | list[str]] is consistently
+        typed before being handed to Workload / FioCommand (which expect
+        dict[str, str]).
+        """
+        normalised: WorkloadType = {}
+        for key, value in options.items():
+            if isinstance(value, list):
```
gitkenan, line 188: This only handles 1 level of nesting (list vs non-list).

If an option is a nested dict (e.g., `auth: {config: ..., s3_session_token: ...}` in elbencho or prefill: {blocksize: '4M', numjobs: 1}), f"{value}" stringifies the entire dictionary into "{'blocksize': '4M', 'numjobs': 1}", which could corrupt nested configuration structures

gitkenan (commented):
Two small comments about the approach here, but overall looks great

workloads/workloads.py
harriscr: Yep, will fix

workloads/workloads.py
harriscr: By the time the code gets here there should be no nested options. prefill is not included in the workloads section, it is handled by the separate prefill() method, which is called at line 75 in cbt.py. Prefill should be a property of the benchmark, not the Workload. We could change this in the future, but then that change would also have to consider the changes required here to handle the new format correctly.

The type to be passed is a dict[str, WorkloadType] where WorkloadType is dict[str, Union[str, list[str]]]

The elbencho code needs to honour this type when calling _create_configurations()

gitkenan (approved):
LGTM!

PR #369, opened 2026-09-22
Updates to requirements.txt

harriscr:
When running with python 3.12 on Ubuntu through teuthology there is a missing import in requirements.txt.

ModuleNotFoundError: No module named 'typing_extensions'

The module is a standard python one that can be installed via pip, so add it to requirements.txt


mohant-ibm (approved):
Thanks!

